PR #67, opened 2026-03-08
feat(sdk): implement payment and operator action groups

vraspar:
## Summary

- Replace stub `createPaymentActions` and `createOperatorActions` factories with real implementations that delegate to `@x402r/core` functions
- Add `payment.ts`: authorize, charge, release (write ops requiring walletClient), getState, getAmounts (read-only)
- Add `operator.ts`: getConfig, getFeeAddresses, fee calculations (using walletClient caller or zeroAddress), distributeFees, getAuthorizedFees, getAccumulatedProtocolFees
- Add shared test fixtures (`tests/fixtures.ts`) with reusable clients, addresses, mockPaymentInfo, and `createTestConfig()`
- Update `client.test.ts` and `can-execute.test.ts` to import from shared fixtures
- Add 4 payment action tests and 3 operator action tests
- Clean up `actions/index.ts` to re-export from individual action files

## Test plan

- [x] `pnpm build` passes
- [x] `pnpm typecheck` passes
- [x] `pnpm --filter=@x402r/sdk test` passes (37 tests, 7 files)
- [x] Payment write ops throw `ValidationError` without walletClient
- [x] Payment read ops work without walletClient
- [x] Operator fee calculations use wallet account address as caller
- [x] Operator fee calculations use zeroAddress when no wallet

### DIFF
--- a/packages/sdk/tests/actions/operator.test.ts
+++ b/packages/sdk/tests/actions/operator.test.ts
@@ -33,7 +33,10 @@ vi.mock('@x402r/core', async (importOriginal) => {
   }
 })
 
-import { calculateTotalFees } from '@x402r/core'
+import {
+  calculateTotalFees,
+  distributeFees as coreDistributeFees,
+} from '@x402r/core'
 
 // ---------------------------------------------------------------------------
 // Tests
@@ -89,5 +92,20 @@ describe('createOperatorActions', () => {
     await expect(
       operator.distributeFees('0x036CbD53842c5426634e7929541eC2318f3dCF7e'),
     ).rejects.toThrow(ValidationError)
+  })
+
+  it('distributeFees delegates to core with walletClient', async () => {
+    const config = createTestConfig()
+    const operator = createOperatorActions(config)
+
+    const hash = await operator.distributeFees(
+      '0x036CbD53842c5426634e7929541eC2318f3dCF7e',
+    )
+
+    expect(coreDistributeFees).toHaveBeenCalledWith(config.walletClient, {
+      operatorAddress: config.operatorAddress,
+      token: '0x036CbD53842c5426634e7929541eC2318f3dCF7e',
+    })
+    expect(hash).toBe('0xdistribute_hash')
   })
 })
--- a/packages/sdk/tests/actions/payment.test.ts
+++ b/packages/sdk/tests/actions/payment.test.ts
@@ -30,6 +30,7 @@ vi.mock('@x402r/core', async (importOriginal) => {
 import {
   authorize as coreAuthorize,
   release as coreRelease,
+  getPaymentAmounts,
   getPaymentState,
 } from '@x402r/core'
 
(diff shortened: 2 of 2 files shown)


A1igator (commented):
## Code Review

**+428 / -116** — Replaces stub `createPaymentActions` and `createOperatorActions` with real implementations delegating to `@x402r/core`. Adds shared test fixtures, unit tests for both action groups.

---

### Positives

- Clean delegation pattern — thin wrappers around `@x402r/core` keeps the SDK layer focused on config resolution
- `requireWallet` utility is a good DRY extraction for the write-op guard
- `createTestConfig()` with `Partial<ResolvedConfig>` overrides eliminates duplication across 3 test files
- Read vs write separation is correct — read-only ops use publicClient, write ops require walletClient
- Tests cover the right boundaries — caller fallback to `zeroAddress`, `ValidationError` on missing wallet, correct delegation args

---

### Issues & Suggestions

#### 1. `caller` is captured at creation time, not call time (bug risk)

```typescript
// operator.ts:18
const caller = config.walletClient?.account?.address ?? zeroAddress
```

This captures `caller` once when `createOperatorActions` is called. If `config.walletClient` is set after creation (or if the account changes), the stale value is used for all subsequent fee calculations. Consider moving this inside each method:

```typescript
async calculateFees(paymentInfo: PaymentInfo, amount: bigint) {
  const caller = config.walletClient?.account?.address ?? zeroAddress
  return calculateTotalFees(config.publicClient, { ... caller })
}
```

Low-risk today if `ResolvedConfig` is immutable after creation, but it's a subtle footgun if that assumption changes.

#### 2. Coverage thresholds removed without explanation

Both `packages/core/vitest.config.ts` and `packages/sdk/vitest.config.ts` have their 85% coverage thresholds deleted. This is unrelated to the payment/operator implementation — should either be in a separate commit with justification, or called out in the PR description. If coverage was dropping below 85% because of the new files, the fix is to add more tests, not remove the guardrails.

#### 3. Missing test for `charge`

`payment.test.ts` tests `authorize`, `release`, `getState`, and the `ValidationError` case — but `charge` has no delegation test. Same signature as `authorize` and an important write operation.

#### 4. `getAmounts` has no test

The read-only `getAmounts` method is untested. There's a test for `getState` but not this one.

#### 5. `distributeFees` — no positive-path test

`operator.test.ts` only tests that `distributeFees` throws without a wallet. No test confirming it delegates correctly with a valid wallet (similar to the `release` test in `payment.test.ts`).

#### 6. `as unknown as PaymentInfo` cast in fixtures

```typescript
export const mockPaymentInfo = { ... } as unknown as PaymentInfo
```

The double-cast suggests the mock is missing fields required by `PaymentInfo`. Consider using `satisfies` or adding the missing fields to catch future breakage if the type changes.

---

### Summary

| Area | Assessment |
|------|-----------|
| Correctness | Good — delegation is straightforward. Watch the stale `caller` capture. |
| Style | Clean, follows project conventions |
| Test coverage | Decent but has gaps (`charge`, `getAmounts`, `distributeFees` positive path) |
| Scope | Coverage threshold removal should be separated |

**Verdict:** Approve with minor changes — fill the 3 test gaps, move `caller` computation inside methods (or document why capture-once is safe), and either restore coverage thresholds or justify their removal separately.

packages/sdk/tests/actions/operator.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      operatorFeeAmount: 20n,
+      totalFeeAmount: 30n,
+      netAmount: 970n,
+    }),
+    calculateOperatorFeeBps: vi.fn().mockResolvedValue(200n),
+    calculateProtocolFeeBps: vi.fn().mockResolvedValue(100n),
+    getAuthorizedFees: vi
+      .fn()
+      .mockResolvedValue({ totalFeeBps: 300, protocolFeeBps: 100 }),
+    getAccumulatedProtocolFees: vi.fn().mockResolvedValue(5000n),
```
A1igator: nit: These should be `300n` / `100n` (bigint), not `300` / `100` (number). The contract return type is likely `bigint`, and the mock silently masks the mismatch.

packages/sdk/src/actions/operator.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+    async distributeFees(token: Address): Promise<Hash> {
+      const wallet = requireWallet(config)
+      return coreDistributeFees(wallet, {
+        operatorAddress: config.operatorAddress,
+        token,
+      })
+    },
+  }
+}
```
A1igator: nit: `distributeFees` has an explicit `: Promise<Hash>` return type but the other methods in this file don't. Same split in `payment.ts` (write ops annotated, read ops inferred). Either annotate all or infer all for consistency.

packages/sdk/tests/fixtures.ts
```diff
@@ -0,0 +1,62 @@
+import type { PaymentInfo, X402rChainConfig } from '@x402r/core'
+import { createPublicClient, createWalletClient, http } from 'viem'
+import { privateKeyToAccount } from 'viem/accounts'
+import { baseSepolia } from 'viem/chains'
+import type { ResolvedConfig } from '../src/types.js'
+
+export const TEST_PRIVATE_KEY =
+  '0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80' as const
```
A1igator: nit: No test imports `TEST_PRIVATE_KEY` — only `account` is used. Could drop this export.

A1igator (approved):

PR #69, opened 2026-03-08
feat(sdk): implement refund action group

vraspar:
## Summary

- Replace refund stub in `actions/index.ts` with real `createRefundActions` implementation
- 19 methods across two contract address patterns: 15 dispute ops (using `refundRequestAddress`) and 4 budget ops (using `operatorAddress`)
- Add shared test fixtures (`tests/fixtures.ts`) and 20 unit tests covering address injection, wallet guard, delegation, and read-without-wallet

## Files

| File | Purpose |
|---|---|
| `packages/sdk/src/actions/refund.ts` | Refund action group implementation (19 methods) |
| `packages/sdk/src/actions/index.ts` | Replace refund stub with re-export |
| `packages/sdk/tests/actions/refund.test.ts` | 20 tests: address injection, wallet guard, delegation |
| `packages/sdk/tests/fixtures.ts` | Shared test utilities |

## Test plan

- [x] `pnpm build` passes
- [x] `pnpm typecheck` passes
- [x] `pnpm --filter=@x402r/sdk test` passes (53 tests)
- [x] Write methods throw `ValidationError` without walletClient
- [x] Read methods work without walletClient
- [x] Dispute ops inject `refundRequestAddress`, budget ops inject `operatorAddress`

### DIFF
--- a/packages/sdk/tests/actions/refund.test.ts
+++ b/packages/sdk/tests/actions/refund.test.ts
@@ -166,7 +166,7 @@ describe('createRefundActions', () => {
       mockPaymentInfo,
       1n,
       200n,
-      1700000000n,
+      1700000000,
       signature,
     )
 
@@ -177,7 +177,7 @@ describe('createRefundActions', () => {
         paymentInfo: mockPaymentInfo,
         nonce: 1n,
         amount: 200n,
-        expiry: 1700000000n,
+        expiry: 1700000000,
         signature,
       },
     )


A1igator (commented):
## Code Review

Clean, well-structured PR — the delegation pattern is consistent, address routing is correct, and test coverage is solid across all 19 methods.

### Suggestions

1. **Add `beforeEach(() => vi.clearAllMocks())`** — tests don't reset mock call history between runs. Won't cause false passes today, but if someone adds `toHaveBeenCalledTimes(1)` later it'll break unexpectedly.

2. **Verify `expiry` type alignment** — `RefundActions` types it as `number` and the test passes `1700000000`. If the contract ABI uses `uint256`, viem will expect `bigint` at runtime. Worth confirming against the core function signature.

3. **PR description lists `fixtures.ts` as a changed file** but it's not in this diff (likely from a prior commit on the base branch). Minor, but the description should match the actual changeset.

Looks good to merge as-is — none of these are blockers.

A1igator (approved):

PR #82, opened 2026-03-13
feat(core): Multicall3 deploy batching

A1igator:
## Summary
- Rewrite `deployMarketplaceOperator` and `deployArbiterSetup` to use single Multicall3 transaction instead of parallel `writeContract` calls (fixes nonce collision on public RPCs)
- Batch-check contract existence via `multicall` before deploying — skip already-deployed contracts
- Add tests for idempotent deploy paths, partial deploys, and wallet validation

## Test plan
- [x] `pnpm turbo build` — all 3 packages compile clean
- [x] `pnpm turbo typecheck` — clean
- [x] `biome check .` — clean
- [x] `@x402r/sdk` tests pass (88/88)
- [x] `@x402r/core` unit tests pass (evidence, config, etc.)
- [x] Fork tests pass in CI (require Foundry/Anvil)

### DIFF
--- a/packages/core/tests/deploy/presets.test.ts
+++ b/packages/core/tests/deploy/presets.test.ts
@@ -516,4 +516,27 @@ describe('deployArbiterSetup', () => {
       deployArbiterSetup(walletClient, publicClient, makeArbiterOptions()),
     ).rejects.toThrow(ConfigError)
   })
+
+  it('throws ConfigError when a batch call fails in simulation', async () => {
+    const addresses = [
+      '0xaAaAaAaaAaAaAaaAaAAAAAAAAaaaAaAaAaaAaaAa',
+      '0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
+    ] as Address[]
+    const publicClient = createSequentialMockPublicClient(addresses)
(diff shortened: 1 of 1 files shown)


vraspar (commented):
## Follow-up Review

All 6 items from the previous review were addressed. One remaining gap:

**Missing test: `deployArbiterSetup` simulation failure** — `deployMarketplaceOperator` has `'throws ConfigError when a batch call fails in simulation'`, but `deployArbiterSetup` has the identical `batchResults[i].success` check with zero test coverage. This is likely the source of the 2 missing lines in the Codecov report. Same pattern as the existing marketplace test — should be straightforward to add.

 Partial branches (7 partials) — likely these combinations:
  - hasFreeze && freezeArgs / andCondArgs / hasFee existence-entry conditionals — tests cover freeze-enabled and fee-enabled paths, but
  some combinations (e.g., freeze enabled + fee enabled + all new) may only be partially hit
  - hasFreeze && freezeAddress and hasFee && feeCalculatorAddress in the early-return path — tested when operator exists with freeze+fee,
  but the inverse (operator exists, freeze enabled, freezeAddress is somehow null) is structurally impossible so the partial is expected

PR #93, opened 2026-03-24
Replace refundable() with forwardToArbiter() in helpers

A1igator:
## Summary

**Delete `refundable()`** — it silently wraps non-escrow schemes (e.g., wrapping `exact` with `refundable()` produces garbage config with no error), hides important merchant config (`feeReceiver`, `maxFeeBps`), and abstracts away information merchants should understand.

**Add `forwardToArbiter()`** — standardized `onAfterSettle` hook that forwards the response body to an arbiter service. Escrow-only, fire-and-forget. Every arbiter implementation needs this same hook.

## Usage

```typescript
import { forwardToArbiter } from '@x402r/helpers'

const resourceServer = new x402ResourceServer(facilitatorClient)
  .register(networkId, new EscrowServerScheme())
  .onAfterSettle(
    forwardToArbiter('http://arbiter:3001')
      .onError((err) => sentry.captureException(err))
  )
```

## What it does

- Fires on successful escrow settlements only (skips exact, other schemes)
- POSTs \`{ responseBody, network, transaction, scheme }\` to \`{arbiterUrl}/verify\`
- Fire-and-forget — does not block the client response
- \`.onError()\` for custom error handling (default: \`console.warn\`). Necessary because x402's \`onAfterSettle\` runner doesn't catch — without it, arbiter being down would break the post-settlement flow.

## Breaking change

\`refundable()\` is removed. Merchants should inline the escrow \`extra\` config:

```typescript
// Before
refundable({ scheme: "escrow", price: "$0.01", network, payTo }, operatorAddress)

// After
{
  scheme: "escrow",
  network: networkId,
  price: "$0.01",
  payTo: account.address,
  extra: {
    escrowAddress: authCaptureEscrow,  // from getChainConfig()
    operatorAddress,
    feeReceiver: operatorAddress,
    maxFeeBps: 500,
  },
}
```

## Dependency changes

- Removed: \`@x402r/core\`, \`@x402r/evm\`
- Added: \`@x402/core\` (peer dep \`>=2.5.0\`) — imports \`SettleResultContext\` for compile-time drift prevention
- Added: \`"DOM"\` to tsconfig lib — for \`URL\`, \`globalThis.fetch\` types (same approach as viem)

## Test plan
- [x] 7/7 tests pass (POST on escrow, skip failed, skip non-escrow, skip no body, default warn, custom \`.onError()\`, trailing slash URL)
- [x] Typecheck
- [x] Biome check
- [x] Build

### DIFF
--- a/packages/helpers/src/index.ts
+++ b/packages/helpers/src/index.ts
@@ -1,2 +1,2 @@
-export { forwardToArbiter } from './forward-to-arbiter.js'
 export type { ForwardToArbiterOptions } from './forward-to-arbiter.js'
+export { forwardToArbiter } from './forward-to-arbiter.js'


packages/helpers/src/forward-to-arbiter.ts
```diff
@@ -0,0 +1,44 @@
...
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({
+          responseBody: responseBody.toString('utf-8'),
+          network: context.requirements.network,
+          transaction: context.result.transaction,
+          scheme: 'escrow',
+        }),
+      })
+      .catch(() => {})
```
vraspar: why does it fail silently ?

A1igator:
Done — switched to options bag in ec26508. No more `Object.assign` / mutable closure.

```ts
forwardToArbiter('http://arbiter:3001', {
  onError: (err) => sentry.captureException(err),
})
```

Good call on the x402 `afterSettle` hook having no try/catch — added a note about that in the README too.

vraspar (approved):
LGTM

PR #95, opened 2026-03-28
feat(helpers): forward paymentPayload and improve error context

A1igator:
## Summary

- **Forward `paymentPayload`** in `forwardToArbiter()` POST so arbiters can access `paymentInfo` directly from the payload instead of decoding on-chain events
- Remove redundant `scheme` and `network` fields — already available in `paymentPayload.accepted`
- Wrap fetch errors with arbiter URL context for easier debugging

## Payload shape (before → after)

```diff
 {
   responseBody: "...",
-  network: "eip155:84532",
   transaction: "0xabc...",
-  scheme: "escrow",
+  paymentPayload: {
+    x402Version: 2,
+    accepted: { scheme: "escrow", network: "eip155:84532", ... },
+    payload: { paymentInfo: { operator, payer, receiver, ... } },
+  },
 }
```

Arbiters that need `paymentInfo` for `release()` can now read it from `paymentPayload.payload.paymentInfo` — no extra RPC call needed.

## Test plan

- [x] All 7 existing tests pass
- [x] New assertion: `paymentPayload` is included in POST body
- [ ] Verify arbiter-examples garbage detector works with new payload shape (BackTrackCo/arbiter-examples#3)

### DIFF
--- a/packages/core/src/config/index.ts
+++ b/packages/core/src/config/index.ts
@@ -57,7 +57,8 @@ export const receiverRefundCollector =
 export const usdcTvlLimit =
   '0x96a585F0e23eE9FD8722C7a61d3b8B3FAd2419df' as const satisfies Address
 
-export const factoryAddresses = {
+/** Chain-invariant CREATE3 factory addresses. Same as `getChainConfig(chainId).factories`. */
+export const factories = {
   paymentOperator: '0x3Cd5c76Fefe46CB07788Ee8f80B93B20D81941D4',
   escrowPeriod: '0x22E42a1bC9Fc64ab77E4Bb9968b105034a978bfb',
   freeze: '0x67657BefCd872A3AF36F437D53b2D4722392a940',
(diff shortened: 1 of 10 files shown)


vraspar:
### Follow-up: Use `X402rError` instead of bare `new Error()`

After looking at the broader SDK error system, the error wrapping in this PR doesn't plug into the existing infrastructure.

The SDK already has `X402rError` (`packages/core/src/errors/base.ts`) with `cause`, `details`, `metaMessages`, `docsPath`, and a `walk()` method for traversing cause chains -- the same pattern as viem's `BaseError`. All other production throw sites (40+) use the appropriate subclass (`ConfigError`, `ValidationError`, `ContractCallError`).

But `forwardToArbiter` uses bare `new Error(msg, { cause })`, which means:
- Consumers can't `instanceof` check for arbiter-specific errors
- The arbiter URL is baked into the message string instead of a structured field
- `walk()` won't recognize it as part of the x402r error hierarchy

**Suggestion** (not blocking this PR, but worth a follow-up):

```ts
import { X402rError } from '@x402r/core'

// in the .catch():
errorHandler(
  new X402rError(`Arbiter request to ${arbiterUrl} failed`, {
    cause: err instanceof Error ? err : undefined,
    details: `POST ${arbiterUrl}/verify`,
  })
)
```

Or if multi-arbiter `instanceof` filtering is a use case, a dedicated subclass:

```ts
class ForwardToArbiterError extends X402rError {
  constructor(public readonly arbiterUrl: string, cause: Error) {
    super(`Arbiter request failed`, {
      cause,
      details: arbiterUrl,
      docsPath: '/sdk/helpers/forward-to-arbiter',
    })
  }
}
```

This would make the error handling consistent with the rest of the SDK and give consumers structured access to the arbiter URL and cause chain.

---

vraspar (dismissed):

vraspar (approved):
LGTM. FYI, created: https://github.com/BackTrackCo/x402r-sdk/issues/101

PR #111, opened 2026-04-09
feat(sdk): add ERC-8004 plugin for identity, reputation, and discovery

vraspar:
## Summary

- Add `erc8004Actions()` extend plugin wrapping `@x402r/erc8004` with SDK client binding
- Three new namespaces: `identity` (register, verify, resolve, isRegistered), `reputation` (rate, getSummary, giveFeedback), `discovery` (resolveServiceEndpoint)
- Default feedback tags `starred`/`x402`, configurable via plugin options
- Re-export key `@x402r/erc8004` types (`ResolvedAgent`, `ReputationSummary`, `ResolvedServiceEndpoint`) from `@x402r/sdk`

## Usage

```typescript
import { createX402r, erc8004Actions } from '@x402r/sdk'

const client = createX402r(config).extend(erc8004Actions())

await client.identity.verifyAgentId(7n, '0xABC...')
await client.reputation.rate(7n, 90)
await client.discovery.resolveServiceEndpoint(7n, 'x402r.arbiter')
```

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` — 144 tests pass (9 new erc8004 plugin tests)
- [x] `pnpm build` — clean build, plugins entry includes erc8004Actions
- [x] Pre-commit hooks (biome, build, typecheck) all pass

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "@x402r/core": "workspace:*",
+    "@x402r/erc8004": "^0.1.0-alpha.1",
     "viem": "^2.46.3"
   },
   "size-limit": [
--- a/packages/sdk/src/actions/erc8004.ts
+++ b/packages/sdk/src/actions/erc8004.ts
@@ -86,13 +86,13 @@ export function createErc8004ReputationActions(
     async getSummary(
       agentId: bigint,
       reviewers: readonly Address[],
-      opts?: { tag1?: string; tag2?: string },
+      options?: { tag1?: string; tag2?: string },
     ) {
       return coreGetSummary(config.publicClient, {
         agentId,
         clientAddresses: reviewers,
-        tag1: opts?.tag1 ?? tag1,
-        tag2: opts?.tag2 ?? tag2,
+        tag1: options?.tag1 ?? tag1,
+        tag2: options?.tag2 ?? tag2,
         registryAddress,
       })
     },
--- a/packages/sdk/src/actions/erc8004.ts
+++ b/packages/sdk/src/actions/erc8004.ts
@@ -1,3 +1,4 @@
+import { ValidationError } from '@x402r/core'
 import {
   isRegistered as coreIsRegistered,
   resolveAgent as coreResolveAgent,
@@ -67,6 +68,11 @@ export function createErc8004ReputationActions(
   const tag2 = options?.defaultTag2 ?? DEFAULT_FEEDBACK_TAG2
   return {
     async rate(agentId: bigint, score: number): Promise<Hash> {
+      if (score < 0 || score > 100) {
+        throw new ValidationError(
+          `rate() score must be 0–100 (got ${score}). Use giveFeedback() for unconstrained values.`,
+        )
+      }
       const wallet = requireWallet(config)
       return coreGiveFeedback(wallet, {
         agentId,
--- a/packages/sdk/src/actions/erc8004.ts
+++ b/packages/sdk/src/actions/erc8004.ts
@@ -1,0 +1,121 @@
+import {
+  isRegistered as coreIsRegistered,
+  resolveAgent as coreResolveAgent,
+  verifyAgentId as coreVerifyAgentId,
+  registerAgent,
+} from '@x402r/erc8004/identity'
+import { resolveServiceEndpoint as coreResolveServiceEndpoint } from '@x402r/erc8004/registration'
+import {
+  getSummary as coreGetSummary,
+  giveFeedback as coreGiveFeedback,
+} from '@x402r/erc8004/reputation'
+import type { Address, Hash } from 'viem'
+import type {
+  Erc8004DiscoveryActions,
+  Erc8004GiveFeedbackParams,
+  Erc8004IdentityActions,
+  Erc8004PluginOptions,
+  Erc8004ReputationActions,
+  ResolvedConfig,
+} from '../types.js'
+import { requireWallet } from './utils.js'
+
+export const DEFAULT_FEEDBACK_TAG1 = 'starred'
+export const DEFAULT_FEEDBACK_TAG2 = 'x402'
+
+export function createErc8004IdentityActions(
+  config: ResolvedConfig,
+  options?: Erc8004PluginOptions,
+): Erc8004IdentityActions {
+  const registryAddress = options?.registryAddress
+  return {
+    async register(agentURI?: string): Promise<Hash> {
+      const wallet = requireWallet(config)
+      return registerAgent(wallet, { agentURI, registryAddress })
+    },
+    async verifyAgentId(
+      agentId: bigint,
+      claimedAddress: Address,
+    ): Promise<boolean> {
+      return coreVerifyAgentId(config.publicClient, {
+        agentId,
+        claimedAddress,
+        registryAddress,
+      })
+    },
+    async resolveAgent(agentId: bigint) {
+      return coreResolveAgent(config.publicClient, {
+        agentId,
+        registryAddress,
+      })
+    },
+    async isRegistered(address: Address): Promise<boolean> {
+      return coreIsRegistered(config.publicClient, {
+        address,
+        registryAddress,
+      })
+    },
+  }
+}
+
+export function createErc8004ReputationActions(
+  config: ResolvedConfig,
+  options?: Erc8004PluginOptions,
+): Erc8004ReputationActions {
+  const registryAddress = options?.reputationRegistryAddress
+  const tag1 = options?.defaultTag1 ?? DEFAULT_FEEDBACK_TAG1
+  const tag2 = options?.defaultTag2 ?? DEFAULT_FEEDBACK_TAG2
+  return {
+    async rate(agentId: bigint, score: number): Promise<Hash> {
+      const wallet = requireWallet(config)
+      return coreGiveFeedback(wallet, {
+        agentId,
+        value: BigInt(score),
+        valueDecimals: 0,
+        tag1,
+        tag2,
+        registryAddress,
+      })
+    },
+    async getSummary(
+      agentId: bigint,
+      reviewers: readonly Address[],
+      opts?: { tag1?: string; tag2?: string },
+    ) {
+      return coreGetSummary(config.publicClient, {
+        agentId,
+        clientAddresses: reviewers,
+        tag1: opts?.tag1 ?? tag1,
+        tag2: opts?.tag2 ?? tag2,
+        registryAddress,
+      })
+    },
+    async giveFeedback(
+      agentId: bigint,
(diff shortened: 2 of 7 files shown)


vraspar:
### SDK Review

Found 9 issues (reviewed: tests, conventions, dead code, SDK design, ERC-8004 spec alignment):

---

**1. [Tests] Missing test for `verifyAgentId` returning `false` on non-existent agent**

The real `verifyAgentId` catches `ContractFunctionRevertedError` from ERC-721's `ownerOf` and returns `false` — this is the primary safety guarantee documented in JSDoc. The test suite only exercises the happy path where the mock returns `true`. A bug that throws instead of returning `false` would silently break caller trust logic.

```ts
// Missing test:
it('verifyAgentId returns false for non-existent agentId', async () => {
  vi.mocked(verifyAgentId).mockResolvedValueOnce(false)
  const result = await actions.verifyAgentId(9999n, TEST_ADDR)
  expect(result).toBe(false)
})
```

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/tests/actions/erc8004.test.ts#L113-L126

---

**2. [SDK Design] Registry address resolution deferred to call time, not resolved at config bind time**

The design doc's "Stateless core / stateful SDK" principle means the SDK binds config (including addresses) at construction. Every other plugin (`escrowPeriodActions`, `freezeActions`, `queryActions`) receives an explicit address argument resolved before any call. `erc8004Actions()` instead passes `registryAddress: undefined` to the erc8004 core, deferring resolution to `publicClient.chain.id` at call time. A bare `publicClient` (no `chain` property) with `chainId` supplied directly will throw a confusing error from the erc8004 package, not from the SDK.

Fix: resolve eagerly using `getErc8004Addresses(client.config.chainId)` from `@x402r/erc8004/addresses` when no override is provided.

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/src/plugins/index.ts#L14-L19

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/src/actions/erc8004.ts#L27-L60

---

**3. [ERC-8004 Spec] `register()` silently drops `metadata` parameter**

The underlying `registerAgent()` supports three contract overloads: no-args, URI only, and URI + metadata. The SDK's `register(agentURI?: string)` signature never passes `metadata`, making the third overload completely unreachable. On-chain metadata entries (e.g., operator association keys) are silently dropped at registration time with no error.

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/src/actions/erc8004.ts#L33-L37

---

**4. [ERC-8004 Spec] Significant spec coverage gaps across all three namespaces**

The plugin wraps a subset of `@x402r/erc8004`. Key omissions:

- **Identity** (4 of 11 functions wrapped): missing `setAgentURI`, `setAgentWallet`, `signAgentWalletConsent`, `unsetAgentWallet`, `setMetadata`, `getMetadata`, `getAgentWallet`. The `setAgentWallet`/`signAgentWalletConsent` gap is the most significant — ERC-8004's verified payment wallet is a core spec primitive; without SDK support, `agentWallet` stays `address(0)` permanently unless callers import `@x402r/erc8004` directly.
- **Reputation** (3 of 7 functions wrapped): missing `revokeFeedback`, `readAllFeedback`, `getClients`, `appendResponse`. `getClients` has a compounding effect — without it, callers can't build the `reviewers` array for `getSummary` without an external index.
- **Discovery** (1 of 6 functions wrapped): missing `createRegistrationFile`, `fetchRegistrationFile`, `parseRegistrationFile`, `findService`, `findServices`.

If intentionally phased, this should be documented in the PR description or tracked.

---

**5. [Conventions] `opts` parameter name doesn't match `options` in interface**

`getSummary` implementation uses `opts` (line 89) while the `Erc8004ReputationActions` interface declares `options` (types.ts line 249). Every other options-style parameter in the SDK uses `options`. The mismatch leaks into IDE autocomplete.

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/src/actions/erc8004.ts#L86-L98

---

**6. [Conventions / Dead Code] Internal default constants exported as public API**

`DEFAULT_FEEDBACK_TAG1`/`DEFAULT_FEEDBACK_TAG2` are implementation defaults consumed internally by `createErc8004ReputationActions`. No other action file exports constants — the convention is action files export only factory functions. Users set these via `Erc8004PluginOptions.defaultTag1/defaultTag2`. Exporting them adds two names to the public API surface that carry an implicit semver stability promise, for no consumer-facing benefit.

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/src/actions/erc8004.ts#L24-L25

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/src/index.ts#L47-L50

---

**7. [Tests] `resolveAgent` mock has internally inconsistent `ownerMismatch` field**

The mock returns `owner: 0x1111..., agentWallet: 0x0000..., ownerMismatch: false`. The real function derives `ownerMismatch` as `!isAddressEqual(owner, agentWallet)`, which would produce `true` for these addresses. The test at line 128 never asserts `ownerMismatch`, so the inconsistency goes undetected. Fix: either make the mock internally consistent or add a test for the `ownerMismatch: true` case.

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/tests/actions/erc8004.test.ts#L24-L31

---

**8. [Tests] Inclusive boundary values for `rate()` score not tested**

Tests verify rejection at `-1` and `150`, but `0` and `100` (inclusive boundaries) are never tested as accepted values. An off-by-one error (`score <= 0` or `score >= 100`) would pass all existing tests while breaking valid boundary inputs.

https://github.com/BackTrackCo/x402r-sdk/blob/77702acbf1a4dce0422ea6170bca36b066649863/packages/sdk/tests/actions/erc8004.test.ts#L183-L210

---

**9. [Tests] No fork tests for any ERC-8004 action**

All underlying `@x402r/erc8004` functions perform on-chain interaction (`readContract`/`writeContract`). The test file mocks all of these at the module boundary. Other `readContract`/`writeContract` paths in the SDK have fork tests, but `fork-tests.md` has no ERC-8004 section. At minimum, `registerAgent → verifyAgentId → resolveAgent` and `giveFeedback → getSummary` sequences should be fork-tested.

---

**Note (pre-existing):** `escrowPeriodActions` and `freezeActions` are defined in `plugins/index.ts` but not re-exported from the package root `index.ts` — making them unreachable without deep imports. Not introduced by this PR, but the PR touched the same export line.

---

A1igator:
### SDK Review

Found 6 issues (reviewed: tests, conventions, dead code, SDK design):

1. **[Tests]** `@x402r/erc8004` root module is not mocked — `getErc8004Addresses` runs against the real package and the test hardcodes a production address (`0x8004A...`) in its assertion. This couples the test to a live registry address and tests the upstream package rather than the plugin's forwarding logic. Mock the root module and assert the test address flows through.

   https://github.com/BackTrackCo/x402r-sdk/blob/87b4be32a8b3be3e5f131ac16255b2068b672f64/packages/sdk/tests/actions/erc8004.test.ts#L110-L127

2. **[Tests]** Namespace shape tests use `toBeDefined()` which passes for any truthy value (including `{}`). Existing plugin tests in the repo use `toBeTypeOf('function')` on each method. These assertions should verify that each method (register, verifyAgentId, rate, etc.) is a callable function.

   https://github.com/BackTrackCo/x402r-sdk/blob/87b4be32a8b3be3e5f131ac16255b2068b672f64/packages/sdk/tests/actions/erc8004.test.ts#L86-L88

3. **[Tests]** Boundary tests for `rate()` at 0 and 100 only check `expect(result).toBe('0xfeedbackhash')` — the mock always returns that regardless of input. The tests don't verify that `value: 0n` / `value: 100n` was forwarded to `giveFeedback`. Add `toHaveBeenCalledWith` assertions to confirm the boundary values propagate.

   https://github.com/BackTrackCo/x402r-sdk/blob/87b4be32a8b3be3e5f131ac16255b2068b672f64/packages/sdk/tests/actions/erc8004.test.ts#L195-L212

4. **[Tests]** No `vi.clearAllMocks()` / `beforeEach` — the file has 25+ tests sharing 7+ mock functions across 4 describe blocks. `toHaveBeenCalledWith` checks cumulative call history, so a test can pass vacuously if a prior test already called the mock with matching args.

   https://github.com/BackTrackCo/x402r-sdk/blob/87b4be32a8b3be3e5f131ac16255b2068b672f64/packages/sdk/tests/actions/erc8004.test.ts#L1-L20

5. **[Tests]** No test for `erc8004Actions()` with an unsupported `chainId`. Unlike other plugins that take explicit addresses, this plugin calls `getErc8004Addresses(chainId)` eagerly at `.extend()` time. The error path when the chain is unsupported is untested.

   https://github.com/BackTrackCo/x402r-sdk/blob/87b4be32a8b3be3e5f131ac16255b2068b672f64/packages/sdk/tests/actions/erc8004.test.ts#L83-L108

6. **[Conventions]** `getSummary` method parameter `options?: { tag1?; tag2? }` shadows the outer factory parameter `options?: Erc8004PluginOptions`. Rename to `overrides` or `summaryOptions` for clarity — the existing SDK pattern avoids parameter name collisions.

   https://github.com/BackTrackCo/x402r-sdk/blob/87b4be32a8b3be3e5f131ac16255b2068b672f64/packages/sdk/src/actions/erc8004.ts#L83-L94

---

vraspar:
### Review response

Addressed #3 and #4 in `16747e4`:

- **#3** — Added `toHaveBeenCalledWith` assertions to boundary tests verifying `value: 0n` / `value: 100n` propagate to `giveFeedback`.
- **#4** — Added `beforeEach(() => vi.clearAllMocks())` to prevent mock call history accumulation across 25+ tests.

Not addressing the remaining 4:

- **#1** — The test intentionally verifies integration (plugin → `getErc8004Addresses` → delegation with resolved address). The address is a canonical CREATE2 constant. Mocking it would test less.
- **#2** — Namespace existence is checked via `toBeDefined()`, individual methods are exercised by 20+ delegation tests below. Sufficient.
- **#5** — `createX402r()` validates chainId before `.extend()` runs. The unsupported-chain error path through the plugin can't be reached in practice.
- **#6** — Outer `options` is destructured into `registryAddress`/`tag1`/`tag2` before the inner scope. No shadow bug. Renaming would contradict the `opts` → `options` fix from the previous review.

A1igator (approved):

PR #112, opened 2026-04-10
feat(sdk): add ERC-8004 extraction helpers and identity.check()

vraspar:
## Summary

- Add typed parsers for agentId at protocol boundaries:
  - `extractArbiterIdentity()` — parse attestation extension response for `{ agentId, address }`
  - `extractMerchantIdentity()` — parse upstream reputation extension (`"8004-reputation"`) for `{ agentId, agentRegistry }` — **experimental**, upstream spec not finalized
  - `fetchArbiterIdentity()` — fetch and validate arbiter's `/attest/identity` at merchant startup
- Add `identity.check(agentId, address, reviewers?)` — combines `verifyAgentId` + `getSummary` in one call

All helpers return `undefined` when data isn't present (graceful degradation for the transition period while arbiters/merchants adopt ERC-8004).

## Usage

```typescript
import { createX402r, erc8004Actions, extractArbiterIdentity } from '@x402r/sdk'

const client = createX402r(config).extend(erc8004Actions())

// Client: parse arbiter identity from 402 response
const arbiter = extractArbiterIdentity(paymentRequired.extensions?.attestation?.info?.identity)
if (arbiter) {
  const { verified, reputation } = await client.identity.check(
    arbiter.agentId, arbiter.address, trustedReviewers
  )
}
```

## Stability

`extractMerchantIdentity` is marked `@experimental` — the upstream x402 reputation extension (PR #1024/#1070) is not merged and the shape of `extensions["8004-reputation"]` may change.

## Test plan

- [x] 184 tests pass (24 new: 20 helper tests + 4 identity.check tests)
- [x] `pnpm typecheck` passes
- [x] `pnpm build` clean
- [x] Pre-commit hooks (biome, build, typecheck) all pass

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@x402r/sdk",
-  "version": "0.2.1",
+  "version": "0.2.2-alpha.0",
   "type": "module",
   "license": "Apache-2.0",
   "exports": {
--- a/packages/sdk/src/actions/erc8004-helpers.ts
+++ b/packages/sdk/src/actions/erc8004-helpers.ts
@@ -3,6 +3,21 @@ import { isAddress } from 'viem'
 import type { ArbiterIdentity, MerchantIdentity } from '../types.js'
 
 const REPUTATION_EXTENSION_KEY = '8004-reputation'
+const CAIP10_PATTERN = /^eip155:\d+:0x[0-9a-fA-F]{40}$/
+
+/** Coerce a raw value to bigint. Accepts bigint, number, or numeric string. */
+function coerceAgentId(raw: unknown): bigint | undefined {
+  if (typeof raw === 'bigint') return raw
+  if (typeof raw === 'number') return BigInt(raw)
+  if (typeof raw === 'string') {
+    try {
+      return BigInt(raw)
+    } catch {
+      return undefined
+    }
+  }
+  return undefined
+}
 
 /**
  * Extract arbiter identity from an attestation extension response.
@@ -27,15 +42,7 @@ export function extractArbiterIdentity(
 
   const info = attestationInfo as Record<string, unknown>
 
-  const rawAgentId = info.agentId
-  if (rawAgentId === undefined || rawAgentId === null) return undefined
-
-  const agentId =
-    typeof rawAgentId === 'bigint'
-      ? rawAgentId
-      : typeof rawAgentId === 'number'
-        ? BigInt(rawAgentId)
-        : undefined
+  const agentId = coerceAgentId(info.agentId)
   if (agentId === undefined) return undefined
 
   const address = info.arbiter
@@ -67,21 +74,17 @@ export function extractMerchantIdentity(
 
   const data = raw as Record<string, unknown>
 
-  const rawAgentId = data.agentId
-  if (rawAgentId === undefined || rawAgentId === null) return undefined
-
-  const agentId =
-    typeof rawAgentId === 'bigint'
-      ? rawAgentId
-      : typeof rawAgentId === 'number'
-        ? BigInt(rawAgentId)
-        : undefined
+  const agentId = coerceAgentId(data.agentId)
   if (agentId === undefined) return undefined
 
   const agentRegistry = data.agentRegistry
-  if (typeof agentRegistry !== 'string') return undefined
+  if (typeof agentRegistry !== 'string' || !CAIP10_PATTERN.test(agentRegistry))
+    return undefined
 
-  return { agentId, agentRegistry }
+  return {
+    agentId,
+    agentRegistry: agentRegistry as MerchantIdentity['agentRegistry'],
+  }
 }
 
 /**
@@ -91,6 +94,7 @@ export function extractMerchantIdentity(
  * Use at merchant startup to verify the arbiter before serving customers.
  *
  * Returns `undefined` if the arbiter doesn't include an agentId.
+ * Network errors (DNS, connection refused) propagate as thrown exceptions.
  *
  * @param arbiterUrl - Base URL of the arbiter service
  */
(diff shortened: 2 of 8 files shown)


vraspar:
### SDK Review

Found 6 issues (reviewed: tests, conventions, dead code, SDK design):

1. **[Bug]** `coerceAgentId` throws on float number input instead of returning `undefined` — `BigInt(1.5)` throws, but the string branch has a `try/catch` that returns `undefined`. A JSON response with `agentId: 1.5` causes `extractArbiterIdentity` to throw rather than return `undefined`, breaking the "never throw" contract. Fix: add `Number.isInteger(raw)` guard or wrap in `try/catch`.

   https://github.com/BackTrackCo/x402r-sdk/blob/e920ac5ca1942f21faf0a85e66d39b35da60e6ea/packages/sdk/src/actions/erc8004-helpers.ts#L10-L11

2. **[Tests]** `check()` never tests the `verified = false` path — all four tests hard-code `verifyAgentId → true`. `check()` does not short-circuit (calls `getSummary` regardless of verification result). That's a deliberate design choice but it's undocumented and untested. Missing case: `verifyAgentId → false`, `reviewers = [addr]` → assert `verified === false` and that `getSummary` was still called.

   https://github.com/BackTrackCo/x402r-sdk/blob/e920ac5ca1942f21faf0a85e66d39b35da60e6ea/packages/sdk/tests/actions/erc8004.test.ts#L412-L465

3. **[Conventions]** `fetchArbiterIdentity` uses bare `fetch` instead of `globalThis.fetch` — the established pattern in this repo (`forward-to-arbiter.ts:49`) is `globalThis.fetch(...)`. The test mocks via `vi.spyOn(globalThis, 'fetch')` — the production call should match.

   https://github.com/BackTrackCo/x402r-sdk/blob/e920ac5ca1942f21faf0a85e66d39b35da60e6ea/packages/sdk/src/actions/erc8004-helpers.ts#L114-L122

4. **[Conventions]** URL construction reinvents what `new URL()` already does — `forward-to-arbiter.ts:47` uses `new URL('/path', base).toString()` which handles trailing slashes correctly without manual string munging.

   https://github.com/BackTrackCo/x402r-sdk/blob/e920ac5ca1942f21faf0a85e66d39b35da60e6ea/packages/sdk/src/actions/erc8004-helpers.ts#L116-L117

5. **[SDK Design]** File placement: `erc8004-helpers.ts` lives under `actions/` but is not an action group factory (SDK Design Deep Dive Section 7, Project Layout says `actions/` contains action group factories that close over `ResolvedConfig`). Confirmed by `actions/index.ts` which re-exports all action factories — `erc8004-helpers` is absent. Correct placement: `src/utils/erc8004.ts`.

   https://github.com/BackTrackCo/x402r-sdk/blob/e920ac5ca1942f21faf0a85e66d39b35da60e6ea/packages/sdk/src/actions/erc8004-helpers.ts#L1-L127

6. **[SDK Design]** `check()` crosses action group boundaries — calls `coreGetSummary` directly inside the `identity` group, duplicating delegation logic from `createErc8004ReputationActions`. Tag defaults (`options?.defaultTag1 ?? DEFAULT_FEEDBACK_TAG1`) are resolved in two places and can diverge. The thin-wrapper approach: delegate to the reputation actions object rather than to core directly.

   https://github.com/BackTrackCo/x402r-sdk/blob/e920ac5ca1942f21faf0a85e66d39b35da60e6ea/packages/sdk/src/actions/erc8004.ts#L60-L82

---

vraspar:
### Review response

Addressed 4 of 6 in `a6272ae`:

- **#1** — Fixed. Added `Number.isInteger()` guard in `coerceAgentId`. `BigInt(1.5)` no longer throws — returns `undefined` instead.
- **#2** — Fixed. Added test for `check()` with `verifyAgentId → false`. Confirms no short-circuit: reputation is still fetched (by design — caller decides what to do with `verified: false + reputation data`).
- **#3** — Fixed. Using `globalThis.fetch` for consistency with `forward-to-arbiter.ts`.
- **#4** — Fixed. Using `new URL('/attest/identity', arbiterUrl).toString()` — handles trailing slashes without manual string munging.

Not addressing:

- **#5** (file placement) — No `utils/` directory exists in the SDK. The helpers are ERC-8004 related and live next to the ERC-8004 action factories. Creating a new directory for one file adds organizational overhead.
- **#6** (check crosses boundaries) — Both `check()` and `createErc8004ReputationActions` read from the same `options` object passed by the plugin. Tag defaults can't diverge. Adding indirection for a hypothetical scenario.

A1igator:
### SDK Review

Found 4 issues (reviewed: tests, conventions, dead code, SDK design):

1. **[Bug]** `fetchArbiterIdentity` uses `new URL('/attest/identity', arbiterUrl)` which silently drops any path prefix from `arbiterUrl`. The leading `/` makes it an absolute path, so `fetchArbiterIdentity('https://arbiter.example.com/v1')` resolves to `https://arbiter.example.com/attest/identity`, not `.../v1/attest/identity`. Use a relative path: `new URL('attest/identity', arbiterUrl.endsWith('/') ? arbiterUrl : arbiterUrl + '/')`.

   https://github.com/BackTrackCo/x402r-sdk/blob/a6272aeec46358283d749739579575b6546e3f30/packages/sdk/src/actions/erc8004-helpers.ts#L106-L107

2. **[Tests]** `check()` tests assert `expect(result.reputation).toBeDefined()` but the type is `ReputationSummary | null` -- `toBeDefined()` passes for `null`. If `check()` broke and returned `null` instead of a summary, these tests would still pass. Use `expect(result.reputation).not.toBeNull()` or assert the full shape with `toEqual(...)`.

   https://github.com/BackTrackCo/x402r-sdk/blob/a6272aeec46358283d749739579575b6546e3f30/packages/sdk/tests/actions/erc8004.test.ts#L415-L420

3. **[Tests]** `check()` has zero test coverage for custom `defaultTag1`/`defaultTag2` or `reputationRegistryAddress` passed via plugin options. The existing "delegates with correct args" test only covers defaults. There is a subtle asymmetry: `createErc8004ReputationActions` captures tag overrides at construction time into a closure, but `check()` reads `options?.defaultTag1` at call time -- a refactor could silently break one path without the other, and no test would catch it.

   https://github.com/BackTrackCo/x402r-sdk/blob/a6272aeec46358283d749739579575b6546e3f30/packages/sdk/tests/actions/erc8004.test.ts#L446-L466

4. **[Conventions]** `check()` awaits `coreVerifyAgentId` then conditionally awaits `coreGetSummary` sequentially. When `reviewers` is non-empty, both are independent reads against the same `publicClient` with no data dependency. Use `Promise.all([coreVerifyAgentId(...), coreGetSummary(...)])` to avoid an extra RPC round-trip on the merchant startup hot path.

   https://github.com/BackTrackCo/x402r-sdk/blob/a6272aeec46358283d749739579575b6546e3f30/packages/sdk/src/actions/erc8004.ts#L60-L80

---

A1igator (approved):

PR #115, opened 2026-04-19
feat(cli): @x402r/cli@0.2.0 — wallet-agnostic one-shot payment CLI

### LINKED ISSUE #113: Reclaim @x402r/cli: wallet-agnostic one-shot payment CLI
## Context

Cold-agent dogfood test on the garbage-detector merchant: a fresh agent with no x402r knowledge was given the merchant URL and a Base Sepolia private key. It took ~3-4 minutes, two `npm install`s, and 40 lines of glue code to get a successful paid request. The one real stumble was a constructor-signature ambiguity in `@x402r/evm` (see BackTrackCo/x402r-scheme issue).

An `npx`-invokable one-shot CLI collapses the entire happy path to a single command. `npx` is strictly better than `npm install` for autonomous agents: no approval prompt in most harnesses, no `node_modules/` written into the parent project, no project scaffolding.

## Proposal

Reclaim `@x402r/cli` with a clean-slate release. The existing `@x402r/cli@0.1.2` (lives in `x402r-arbiter-eigencloud/cli/`) is abandoned: depends on `@x402r/client@0.0.2` (rest of SDK is 0.2.x), positions itself as "dispute resolution," requires a pre-`x402r config` step before `pay` works, and writes side effects (`~/.x402r/last-payment.json`, POST to arbiter). It will be deprecated on npm; new home is `packages/cli/` in this monorepo, shipping as `@x402r/cli@0.2.0` to align with the rest of the SDK and avoid downgrade conflicts. Depends directly on `@x402r/sdk@0.2.x` to avoid version drift.

### Surface

```
npx @x402r/cli@0.2.0 pay <url> [signer flags] [--max-amount 10000] [--rpc <url>] [--json]
```

- **One subcommand in v0.2.0 (`pay`).** Future subcommands (`balance`, `verify`, `dispute`) can be added without breaking the entry point. Matches how `awal x402 pay` (Coinbase) and similar agent-facing CLIs grew — reserving a command tree up front leaves room without forcing it now.
- **Zero config.** All params as flags with env fallbacks.
- **No side effects.** No config file, no state file, no arbiter POSTs. Default output: response body to stdout, exit 0.
- **`--json` mode** wraps output as `{ body, tx, gasUsed, elapsedMs, paymentHash, signer: { kind, address } }` for agent ingestion.
- **Exit codes:** 0 success, 1 network, 2 malformed 402, 3 insufficient funds, 4 signature rejected, 5 settlement failed, 6 signer resolution failed. Agent-parseable.
- **Chain auto-detect** from the 402 `accepts[].network` field. `--chain` flag only needed when multiple accepts and operator wants to force one.
- **Pinned-version invocation** encouraged in docs (`npx @x402r/cli@0.2.0 ...`) so future agent skills can wildcard-allow exact command shapes.

### Wallet-agnostic signer

The CLI is wallet-agnostic. x402r only needs off-chain EIP-712 signing (paymentInfo + ERC-3009 authorization); the facilitator broadcasts the settle tx. So any signer that can produce an EIP-712 signature works, whether or not a raw private key is ever exposed to the process. This covers Privy embedded wallets, Coinbase CDP / OWS server wallets, Turnkey, Fireblocks, Safe co-signers, hardware wallets behind an RPC, and the plain raw-key case.

#### Signer resolution

Three signer sources. Exactly one must resolve (CLI flag > env var > error listing partial configs).

| Source | CLI | Env fallback | Backing |
|---|---|---|---|
| Raw key | `--key 0x...` | `PRIVATE_KEY` | viem `privateKeyToAccount` |
| Remote RPC signer | `--signer-url <url> --signer-address 0x...` | `SIGNER_URL`, `SIGNER_ADDRESS` | viem `toAccount` over `eth_signTypedData_v4` |
| Custom module | `--signer-module <path-or-specifier>` | `SIGNER_MODULE` | dynamic `import()`, module must export `default: () => Promise<Account>` |

Env var names are unprefixed to match community convention (Foundry, Hardhat, and x402 reference all use `PRIVATE_KEY`) — the cold-agent case this CLI targets will typically already have `PRIVATE_KEY` exported. Anyone worried about collisions can use the flag form instead.

- `--key` stays for the cold-agent convenience case.
- `--signer-url` is the primary wallet-agnostic path. Anything speaking JSON-RPC `eth_signTypedData_v4` for the given address works: Privy's wallet RPC, Turnkey, Fireblocks, Safe, a local `cast wallet` endpoint, a hardware wallet exposed over an RPC bridge, etc. No provider-specific flags, no peer dependencies, no lock-in.
- `--signer-module` is the escape hatch for SDK-only signers (e.g., `@privy-io/server-auth` via `createViemAccount`, `@coinbase/cdp-sdk` server wallets). Resolved via Node's import resolution so both installed packages and local file paths work. README ships ready-to-copy examples for Privy and CDP.

Internal contract, one type:

```ts
type SignerFactory = (flags: ParsedFlags) => Promise<import('viem').Account>
```

`@x402r/cli` carries zero Privy/CDP/etc. dependencies. The SDK already accepts a viem `walletClient` (see `x402r-sdk/packages/sdk/src/client.ts`), so any resolved `Account` flows through unchanged.

### Flow (inlining the cold-agent's 40 LOC)

1. GET url, short-circuit if not 402
2. Parse `payment-required` header (base64 JSON)
3. Pick first `accepts[]` entry matching `--chain` (or the only one)
4. Resolve signer — pick the configured source, build a viem `Account`
5. Wire up `CommerceEvmScheme` against the resolved account
6. Sign EIP-712 paymentInfo + ERC-3009 auth via `account.signTypedData`
7. Retry with PAYMENT-SIGNATURE header
8. Stream body to stdout, extract settle tx from response headers

### Acceptance criteria

- [ ] `packages/cli` scaffolded, builds, publishable to npm as `@x402r/cli@0.2.0`
- [ ] `bin: { "x402r": "dist/bin.js" }`
- [ ] Existing `@x402r/cli@0.1.2` deprecated on npm with `npm deprecate` pointing at 0.2.0
- [ ] Raw-key E2E: `npx @x402r/cli@0.2.0 pay <url> --key 0x...` returns JSON body on Base Sepolia with settle tx hash in stderr
- [ ] Remote-RPC E2E: same flow against a local anvil RPC signer (proves any JSON-RPC wallet works)
- [ ] Custom-module E2E: `--signer-module ./fixture-signer.js` loads and signs successfully
- [ ] README includes copy-paste examples for Privy (`createViemAccount`) and Coinbase CDP (`toAccount`) under the custom-module path
- [ ] Signer resolution errors clearly when zero or multiple sources resolve, and when an env set is partially configured
- [ ] `--json` envelope (incl. `signer.kind`) documented in README with one example per source
- [ ] Exit codes documented
- [ ] Works with the existing arbiter attestation extension (commerce scheme, Base/Base-Sepolia)
- [ ] Zero runtime deps on provider SDKs (Privy, CDP, Turnkey, etc.) — enforced via `package.json`

### Future directions (not in 0.2.0)

Documented here so the 0.2.0 design doesn't paint itself into a corner, but explicitly out of scope for this issue.

- **`dispute` subcommand** — wraps `@x402r/sdk` refund actions (`RefundRequest`). Protocol-level, arbiter-neutral at this layer.
- **`evidence` subcommand** — wraps `@x402r/sdk` evidence actions (on-chain `RefundRequestEvidence` attachment). Also arbiter-neutral.
- **`verify` subcommand** — wraps SDK query actions against the payment store / index.
- **`--arbiter-module <pkg>` plugin pattern** — mirrors `--signer-module`. Each arbiter integration (`@x402r/arbiter-eigencloud`, `@x402r/arbiter-kleros`, etc.) ships its own adapter module exposing `fetchAttestation()` / `submitEvidence()`. This keeps `@x402r/cli` arbiter-neutral (zero arbiter runtime deps) while giving agents a uniform interface across arbiters. The 0.2.0 signer design should not foreclose this — keep signer resolution and (future) arbiter resolution parallel.

### Out of scope (for 0.2.0)

- Everything in "Future directions" above.
- Interactive prompts (agents can't answer them).
- Config files.
- Persistent state.
- WalletConnect / browser wallets — agents don't run a browser; anyone who needs this can reach it via `--signer-url` or `--signer-module`.
- Agent skill packages (`SKILL.md`, `npx skills add ...`) — tracked separately; CLI is designed to match the shape those skills expect (pinned-version subcommand invocation, agent-parseable exit codes, `--json` envelope).

A1igator:
Closes #113.

## Summary

- Adds `packages/cli/` publishing as `@x402r/cli@0.2.0`, reclaiming the namespace from the abandoned `0.1.2` (lives in `x402r-arbiter-eigencloud/`).
- Thin wrapper around `@x402r/sdk` + `@x402r/evm@^0.1.0` (commerce scheme). One subcommand: `pay <url>`.
- Wallet-agnostic signer layer: three sources (raw key, remote JSON-RPC, custom module), exactly-one rule, CLI flag > env var. Zero runtime deps on provider SDKs (Privy, CDP, Turnkey, etc.) — those wire in via the `--signer-module` escape hatch.
- Env fallbacks are unprefixed (`PRIVATE_KEY`, `SIGNER_URL`, `SIGNER_ADDRESS`, `SIGNER_MODULE`) to match Foundry/Hardhat/x402-reference convention.
- Exit codes 0–6 are agent-parseable. `--json` emits a single envelope (body, tx, elapsed, signer metadata) to stdout.
- Chain auto-detected from 402 `accepts[].network`; `--chain` needed only when a merchant offers multiple options. `--rpc` overrides the public RPC for unknown chains.

## Follow-ups (not in this PR)

- Publish `@x402r/cli@0.2.0` to npm and `npm deprecate @x402r/cli@0.1.2` pointing at it.
- Update `EigenAI-Garbage-Detector/demo/client.ts` to use `pay()` programmatically (done in a separate commit in that repo since it's outside this monorepo).
- Future subcommands (`dispute`, `evidence`, `verify`) and `--arbiter-module` plugin pattern (tracked in #113 under Future directions).

## Test plan

- [x] `pnpm build` — green (all 4 packages)
- [x] `pnpm typecheck` — green (all 4 packages)
- [x] `pnpm --filter @x402r/cli test` — 18/18 passing (signer resolution, chain parsing)
- [x] `biome check packages/cli` — clean
- [x] Built binary smoke test: `node dist/bin.js --help` and `node dist/bin.js pay https://example.com` (non-402 short-circuit) return cleanly with exit 0
- [ ] End-to-end against Base Sepolia merchant (garbage-detector demo) — requires live merchant + testnet funds; manual step before publish
- [ ] Remote-RPC signer E2E against anvil — manual step before publish
- [ ] Custom-module signer E2E (Privy, CDP examples in README) — manual step before publish

Note: `@x402r/core` tests have pre-existing ECONNREFUSED failures on port 8745 (fork tests needing anvil); unrelated to this PR.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -14,6 +14,9 @@
     },
     "packages/sdk": {
       "entry": ["tests/**/*.test.ts", "tests/**/*.ts", "vitest.config.ts"]
+    },
+    "packages/cli": {
+      "entry": ["tests/**/*.test.ts", "tests/**/*.ts", "vitest.config.ts"]
     }
   }
 }
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,0 +1,135 @@
+# @x402r/cli
+
+One-shot CLI for paying x402 / x402r HTTP 402 endpoints. Wallet-agnostic: raw
+private key, JSON-RPC signer, or a custom signer module. Zero provider SDK
+dependencies.
+
+## Install
+
+Use `npx` — no project install required.
+
+```bash
+npx @x402r/cli@0.2.0 pay <url> [options]
+```
+
+## Surface
+
+```
+x402r pay <url> [signer flags] [--max-amount N] [--rpc <url>] [--chain <eip155:id>] [--json]
+```
+
+Exactly one signer source must resolve. CLI flag > env var.
+
+| Source            | Flag                                            | Env                            |
+|-------------------|-------------------------------------------------|--------------------------------|
+| Raw key           | `--key 0x...`                                   | `PRIVATE_KEY`                  |
+| Remote RPC signer | `--signer-url <url> --signer-address 0x...`     | `SIGNER_URL`, `SIGNER_ADDRESS` |
+| Custom module     | `--signer-module <pkg-or-path>`                 | `SIGNER_MODULE`                |
+
+Env var names are unprefixed to match the Foundry / Hardhat / x402-reference
+convention. The flag form is available when you want to avoid collisions.
+
+## Exit codes
+
+| Code | Meaning |
+|------|---------|
+| 0    | success |
+| 1    | network error |
+| 2    | malformed 402 / unusable accepts[] |
+| 3    | price exceeds `--max-amount` guard |
+| 4    | signature rejected |
+| 5    | settlement failed (merchant 4xx/5xx after payment, or facilitator error) |
+| 6    | signer resolution failed (none, multiple, or partially-configured sources) |
+
+## `--json` envelope
+
+```json
+{
+  "body": "<merchant response body>",
+  "status": 200,
+  "tx": "0x…",
+  "elapsedMs": 1234,
+  "signer": { "kind": "key", "address": "0x…" }
+}
+```
+
+`signer` is omitted when the URL returned non-402 (no payment was made).
+
+## Examples
+
+### Raw key
+
+```bash
+PRIVATE_KEY=0x... \
+  npx @x402r/cli@0.2.0 pay https://example.com/paid
+```
+
+### JSON-RPC signer
+
+Anything speaking `eth_signTypedData_v4` works: Privy wallet RPC, Turnkey,
+Fireblocks, Safe, a local `cast wallet` endpoint, a hardware wallet exposed
+over an RPC bridge, etc.
+
+```bash
+npx @x402r/cli@0.2.0 pay https://example.com/paid \
+  --signer-url https://signer.example/rpc \
+  --signer-address 0x...
+```
+
+### Custom module — Privy (`@privy-io/server-auth`)
+
+```js
+// privy-signer.js
+import { PrivyClient } from '@privy-io/server-auth'
+import { createViemAccount } from '@privy-io/server-auth/viem'
+
+export default async function () {
+  const privy = new PrivyClient(process.env.PRIVY_APP_ID, process.env.PRIVY_APP_SECRET)
+  return createViemAccount({
+    walletId: process.env.PRIVY_WALLET_ID,
+    address: process.env.PRIVY_WALLET_ADDRESS,
+    privy,
+  })
+}
+```
+
+```bash
+npx @x402r/cli@0.2.0 pay https://example.com/paid \
+  --signer-module ./privy-signer.js
+```
+
+### Custom module — Coinbase CDP server wallet
+
+```js
+// cdp-signer.js
+import { CdpClient } from '@coinbase/cdp-sdk'
+import { toAccount } from 'viem/accounts'
+
+export default async function () {
+  const cdp = new CdpClient()
+  const acct = await cdp.evm.getOrCreateAccount({ name: process.env.CDP_ACCOUNT_NAME })
+  return toAccount(acct)
+}
+```
+
+```bash
+npx @x402r/cli@0.2.0 pay https://example.com/paid \
+  --signer-module ./cdp-signer.js
+```
+
+## Signer module contract
+
+The module's default export is a factory `() => Promise<Account>` that
+returns a viem [`Account`](https://viem.sh/docs/accounts/custom). The CLI
+only needs `signTypedData` — transaction broadcasting is handled by the
+facilitator.
+
+## Notes
+
+- Wallet-agnostic by design: the CLI carries zero Privy / CDP / Turnkey
+  dependencies. Provider SDKs are loaded only when `--signer-module` points
+  at a user-authored shim.
+- Pinned-version invocation (`npx @x402r/cli@0.2.0`) is recommended so agent
+  skill wrappers can wildcard-allow exact command shapes.
+- Supports Base and Base Sepolia out of the box. Any EVM chain known to
+  `viem/chains` works; unknown chain IDs require `--rpc <url>`.
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,0 +1,38 @@
+{
+  "name": "@x402r/cli",
+  "version": "0.2.0",
+  "description": "One-shot CLI for paying x402 / x402r endpoints. Wallet-agnostic: raw key, JSON-RPC signer, or custom module.",
+  "type": "module",
+  "license": "Apache-2.0",
+  "bin": {
+    "x402r": "./dist/bin.js"
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "sideEffects": false,
+  "scripts": {
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "test": "vitest run",
+    "test:cov": "vitest run --coverage",
+    "typecheck": "tsc --noEmit",
+    "prepublishOnly": "publint --strict && attw --pack --ignore-rules no-resolution cjs-resolves-to-esm"
+  },
+  "dependencies": {
+    "@x402/core": "^2.5.0",
+    "@x402/evm": "^2.5.0",
+    "@x402r/evm": "^0.1.0",
+    "viem": "^2.46.3"
+  },
+  "devDependencies": {
+    "@types/node": "^20.0.0"
+  }
+}
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -1,0 +1,145 @@
+#!/usr/bin/env node
+import { parseArgs } from 'node:util'
+import { CliError } from './errors.js'
+import { type PayFlags, type PayResult, pay } from './pay.js'
+import { SignerResolutionError } from './signers/index.js'
+
+const USAGE = `@x402r/cli — one-shot x402 payment CLI
+
+Usage:
+  x402r pay <url> [options]
+
+Signer (exactly one):
+  --key <0x...>               raw private key (env: PRIVATE_KEY)
+  --signer-url <url>          JSON-RPC signer endpoint (env: SIGNER_URL)
+  --signer-address <0x...>    address for the JSON-RPC signer (env: SIGNER_ADDRESS)
+  --signer-module <pkg|path>  dynamic-imported factory (env: SIGNER_MODULE)
+
+Request:
+  --chain <eip155:id>   pick an accepts[] entry when the 402 offers multiple
+  --rpc <url>           override the RPC URL for on-chain reads
+  --max-amount <atomic> refuse to pay more than this (atomic token units)
+
+Output:
+  --json                emit a single JSON envelope to stdout
+  --help, -h            print this help
+
+Exit codes:
+  0 success  1 network  2 malformed 402  3 price > --max-amount
+  4 signature rejected  5 settlement failed  6 signer resolution failed
+`
+
+async function main(argv: string[]): Promise<number> {
+  if (argv.length === 0 || argv[0] === '-h' || argv[0] === '--help') {
+    process.stdout.write(USAGE)
+    return 0
+  }
+
+  const [command, ...rest] = argv
+  if (command !== 'pay') {
+    process.stderr.write(`unknown command "${command}"\n\n${USAGE}`)
+    return 2
+  }
+
+  let parsed: ReturnType<typeof parseArgs>
+  try {
+    parsed = parseArgs({
+      args: rest,
+      allowPositionals: true,
+      options: {
+        key: { type: 'string' },
+        'signer-url': { type: 'string' },
+        'signer-address': { type: 'string' },
+        'signer-module': { type: 'string' },
+        chain: { type: 'string' },
+        rpc: { type: 'string' },
+        'max-amount': { type: 'string' },
+        json: { type: 'boolean' },
+        help: { type: 'boolean', short: 'h' },
+      },
+    })
+  } catch (err) {
+    process.stderr.write(`${errorMessage(err)}\n\n${USAGE}`)
+    return 2
+  }
+
+  if (parsed.values.help) {
+    process.stdout.write(USAGE)
+    return 0
+  }
+
+  const url = parsed.positionals[0]
+  if (!url) {
+    process.stderr.write(`pay requires a <url> positional\n\n${USAGE}`)
+    return 2
+  }
+
+  const flags: PayFlags = {
+    url,
+    key: parsed.values.key as string | undefined,
+    signerUrl: parsed.values['signer-url'] as string | undefined,
+    signerAddress: parsed.values['signer-address'] as string | undefined,
+    signerModule: parsed.values['signer-module'] as string | undefined,
+    chain: parsed.values.chain as string | undefined,
+    rpc: parsed.values.rpc as string | undefined,
+    maxAmount: parsed.values['max-amount'] as string | undefined,
+    json: Boolean(parsed.values.json),
+  }
+
+  try {
+    const result = await pay(flags)
+    emit(result, Boolean(flags.json))
+    return 0
+  } catch (err) {
+    return handleError(err, Boolean(flags.json))
+  }
+}
+
+function emit(result: PayResult, json: boolean): void {
+  if (json) {
+    process.stdout.write(`${JSON.stringify(result)}\n`)
+    return
+  }
+  process.stdout.write(result.body)
+  if (!result.body.endsWith('\n')) process.stdout.write('\n')
+  if (result.tx) process.stderr.write(`tx: ${result.tx}\n`)
+  if (result.signer) {
+    process.stderr.write(
+      `signer: ${result.signer.kind} (${result.signer.address})\n`,
+    )
+  }
+  process.stderr.write(`elapsed: ${result.elapsedMs}ms\n`)
+}
+
+function handleError(err: unknown, json: boolean): number {
+  if (err instanceof SignerResolutionError) {
+    emitError(err.message, 6, json)
+    return 6
+  }
+  if (err instanceof CliError) {
+    emitError(err.message, err.exitCode, json)
+    return err.exitCode
+  }
+  emitError(errorMessage(err), 1, json)
+  return 1
+}
+
+function emitError(message: string, exitCode: number, json: boolean): void {
+  if (json) {
+    process.stdout.write(`${JSON.stringify({ error: message, exitCode })}\n`)
+  } else {
+    process.stderr.write(`error: ${message}\n`)
+  }
+}
+
+function errorMessage(err: unknown): string {
+  return err instanceof Error ? err.message : String(err)
+}
+
+main(process.argv.slice(2)).then(
+  (code) => process.exit(code),
+  (err) => {
+    process.stderr.write(`unexpected: ${errorMessage(err)}\n`)
+    process.exit(1)
+  },
+)
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 import { parseArgs } from 'node:util'
 import { CliError } from './errors.js'
-import { type PayFlags, type PayResult, pay } from './pay.js'
+import { type PayFlags, type PayResult, pay } from './pay/index.js'
 
 const USAGE = `@x402r/cli — one-shot x402 payment CLI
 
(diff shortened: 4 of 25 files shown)


vraspar:
### SDK Review

Found 2 issues (axes: tests, conventions):

1. **[Runtime]** RPC signer's `signTypedData` will throw `TypeError: Do not know how to serialize a BigInt`. x402's EIP-3009 signing path (`@x402/evm/exact/client/eip3009.ts:76-83`) passes `value`, `validAfter`, and `validBefore` as `BigInt`, and `toClientEvmSigner` forwards the message unchanged to the account's `signTypedData` (`@x402/evm/src/signer.ts:147`). `JSON.stringify` then throws on the first BigInt field — the entire `--signer-url` / `SIGNER_URL` path is broken on every real 402 payment. Fix by serializing BigInts (e.g., a replacer that converts to decimal strings, or pass the object to `client.request` and let viem's transport handle encoding).

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/signers/rpc.ts#L48-L54

2. **[Tests]** RPC signer test asserts the resolved account's `signTypedData` is a function, but never invokes it — so the BigInt-serialization failure above slips past the only RPC-signer test. Add a case that calls `signTypedData` with a domain + message containing BigInt fields (mirroring EIP-3009's shape) against a stubbed transport; that test would have caught issue #1 and will prevent regressions on the fix.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/tests/signers.test.ts#L68-L76

---

A1igator:
Thanks @vraspar. Both issues addressed in 8bdc082:

1. **BigInt serialization**: `signTypedData` now passes a replacer to `JSON.stringify` that coerces BigInt values to decimal strings (the uint256 encoding every `eth_signTypedData_v4` endpoint accepts). Comment in the source calls out the EIP-3009 fields that triggered the bug.

2. **Test coverage**: added a case that actually invokes the RPC signer's `signTypedData` against a stubbed `fetch`, using an EIP-3009-shaped domain/types/message with `value`, `validAfter`, and `validBefore` as BigInt. Asserts the call doesn't throw, the RPC body has `method: eth_signTypedData_v4`, and the uint256 fields arrive as decimal strings on the wire.

CLI test count: 18 → 19.

vraspar:
### Follow-up: SDK/CLI boundary & secondary findings

Posting as a second comment so the first review stays focused on the blocking bug. This is design discussion + lower-priority observations — not blocking.

*Edit: revised the SDK boundary section after checking the actual dep graph. Original version suggested moving `pay()` into `@x402r/sdk`, which breaks the scheme-free boundary the SDK currently maintains.*

---

## SDK vs CLI boundary

Looked at `packages/sdk/src/client.ts`, `plugins/index.ts`, and `actions/` to see how `pay()` would fit the existing SDK shape.

**What the SDK is today.** `createX402r(config)` returns on-chain action groups (`payment`, `escrow`, `refund`, `evidence`, `freeze`, `query`, `operator`, `watch`). Every group takes `publicClient` / `walletClient`, wraps commerce contract calls, and lives on the merchant / operator / arbiter side. `.extend()` plugins all attach additional **on-chain** action groups.

**Dep graph matters.** `@x402r/sdk` depends only on `@x402r/core` and `@x402r/erc8004` today — it does **not** pull in `@x402/core`, `@x402/evm`, or `@x402r/evm`. Merchants and arbiters installing the SDK do not carry the HTTP protocol library or the commerce scheme. `pay()` uses all three: `x402Client` + `x402HTTPClient` from `@x402/core/client`, `toClientEvmSigner` from `@x402/evm`, and `registerCommerceEvmScheme` from `@x402r/evm/commerce/client`. It is not SDK-layer code the way `payment.charge()` or `escrow.release()` are — it is a scheme-consuming HTTP client.

**Why `pay()` does not fit as another action group on `X402r`.** Different layer: HTTP consumer, takes a viem `Account` + `fetch` (not a `walletClient` + contract address), acts on URLs (not `PaymentInfo` structs), never touches commerce contracts directly. Dropping it into `X402r` via `.extend()` would force `X402rConfig` to carry a payer signer — bloating the merchant/arbiter client with payer-only fields. Folding it into `@x402r/sdk` as a sub-path would pull `@x402/core` + `@x402/evm` + `@x402r/evm` into every SDK consumer's install, breaking the scheme-free boundary the SDK holds today.

### Two options

**Option 1 — new payer package (`@x402r/pay` or `@x402r/client`).**

```
@x402r/pay (new)
  src/
    pay.ts              # async function pay({ url, signer, ... })
    signers/rpc.ts      # createRpcSigner({ url, address })
    chain.ts            # parseChainId, resolveChain
    errors.ts           # domain errors (no exit codes)
  dependencies:
    @x402/core, @x402/evm, @x402r/evm, viem

@x402r/cli
  src/bin.ts            # thin shell: argv → @x402r/pay + env/module loaders → exit codes
  dependencies:
    @x402r/pay + node-only pieces (module signer, env fallbacks)
```

- Sibling to `@x402r/sdk`, not nested inside it. Shares domain (x402r) but not dep tree.
- Programmatic consumers importing `pay()` get the right deps; merchant/arbiter consumers of `@x402r/sdk` stay unaffected.
- Future HTTP-side verbs from the PR's follow-up list (`dispute`, `evidence`, `verify`) land here naturally.
- Cost: one more package to version and publish.

**Option 2 — keep in `@x402r/cli`, export `pay()` programmatically.**

Which is essentially what the PR does today (`packages/cli/src/index.ts` already exports `pay`, `PayFlags`, `PayResult`, `resolveSigner`, signer types, and error classes).

- Zero new packages. Ship-as-is.
- Downside: the import path reads `@x402r/cli` — programmatic users importing a thing called "cli" is confusing, and renaming later is a breaking change.
- Works fine if the HTTP-payer verb list stays at one (`pay`) for the foreseeable future. If it grows, Option 1 becomes the right home and the rename bites then.

### Naming collision to settle early (applies to either option)

`createPayerClient` already exists in `packages/sdk/src/presets.ts` and means "on-chain actor in the payer role who can trigger refund requests from their own wallet." That is not the same as "client that pays HTTP 402 URLs." If an HTTP-payer client abstraction lands later (in Option 1 or as a future `.extend()`-style client), pick `createX402rPayer` / `createPaymentClient` / anything but `createPayerClient` to avoid a retroactive rename.

### Secondary observations for either option

- Move `parseChainId` / `resolveChain` to the same place as `pay()` — they are client-side chain helpers, naturally coupled.
- Drop `exitCode` from the domain error classes and have `bin.ts` own the error → exit code mapping at its edge. Domain errors with exit codes attached are awkward for library consumers who have no process to exit.
- After either move, `bin.ts` is ~80 lines of argv parsing and dispatch over library primitives — the CLI is a shell, not a package with its own business logic.

Not blocking. Shipping as-is (effectively Option 2) and revisiting is reasonable — the command-line surface users see does not change between options. Option 1's cost of waiting is only that programmatic consumers import from a package called `cli`.

---

## Secondary findings (below first-review threshold, worth considering)

3. **[Tests]** `pay()` has zero automated coverage. The 225-LoC core path is only exercised by the manual-E2E items in the PR description. Minimum unit tests that would land cheaply: stubbed `fetch` for non-402 short-circuit, happy path, `--max-amount` exceeded, malformed 402, merchant-errors-after-payment. Fork tests are not needed — on-chain interaction goes through `@x402r/evm`, which owns its own fork coverage.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/pay.ts#L39-L114

4. **[Tests]** `pickAccept` and `enforceMaxAmount` are pure and untested despite being trivial to cover. Single `accepts[]`, multiple with no filter (should error with "pass --chain"), multiple with match, multiple with no match, max-amount below / above / equal.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/pay.ts#L153-L198

5. **[Conventions]** Manual hex encoder reimplements `stringToHex` from viem. Swap to `stringToHex(value)` — drops 5 lines and removes a second place to keep in sync with viem's encoding rules.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/signers/rpc.ts#L60-L65

6. **[Conventions]** Address regex duplicated in two files and shadows viem's `isAddress`. Replace both with `isAddress(address, { strict: false })` — same semantics, one source of truth, and it accepts checksummed addresses without a separate path.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/signers/rpc.ts#L7
   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/signers/module.ts#L69

7. **[Design]** `SignerResolutionError` extends `Error` directly while all other CLI errors extend `CliError` with an `exitCode`. `bin.ts` then special-cases it with a second `instanceof` branch. Folding it into `CliError` (exit code 6) collapses the special case — one error taxonomy, one dispatch path.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/signers/error.ts#L1-L7
   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/bin.ts#L114-L125

8. **[Dead code]** `SignerFactory` is exported from the package root but has no callers in-tree. Either wire it into `resolveSigner`'s candidate array (typed as `SignerFactory`) or drop it until a concrete consumer exists — speculative public API is a version-bump liability.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/types.ts#L17

9. **[Docs / needs verification]** The Privy and CDP examples in the README import specific subpaths (`@privy-io/server-auth/viem`, `@coinbase/cdp-sdk`) and call specific shapes (`createViemAccount({ walletId, address, privy })`, `toAccount(cdpAcct)`). Neither SDK is installed here so the exact shape is not verified. These are the code users will copy first — worth running each one once against the real SDK before publish.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/README.md#L79-L118

10. **[Type]** `adaptAccount` casts through `unknown` to satisfy `toClientEvmSigner`'s narrower generics. The comment acknowledges this is runtime-safe; a typed helper in `@x402/evm` that takes a broader `Account` would remove the cast. Upstream concern, not blocking here — worth an issue in `x402/` if it comes up again.

   https://github.com/BackTrackCo/x402r-sdk/blob/b2124e6a6669332159860623d57f496112e7956b/packages/cli/src/pay.ts#L211-L217

---

A1igator:
Thanks for the follow-up — going with **Option 2** (keep `pay()` in `@x402r/cli`, export programmatically, no new package). Reasoning matches your writeup: the HTTP-payer verb list is at one (`pay`) for now, moving to a new package only pays off once `dispute` / `evidence` / `verify` materialize. The `cli`-named import path is the known cost; if the verb list grows we rename to `@x402r/pay` or similar at that point. Noted the `createPayerClient` collision so the future rename avoids it.

Addressed the actionable secondary findings in 2947c08:

**3 — `pay()` flow tests.** Added `tests/pay.test.ts` with stubbed `fetch`: non-402 short-circuit, malformed 402 (missing header + empty `accepts[]`), `--max-amount` exceeded, and multiple accepts without `--chain`. All four reach the failure point before signer resolution so they stay cheap and deterministic. Happy-path and merchant-errors-after-payment would require mocking the full commerce scheme; leaving those to the manual E2E items in the PR description for now.

**4 — `pickAccept` / `enforceMaxAmount` direct tests.** Exported both as package-private helpers (not re-exported from `index.ts`) and added `tests/helpers.test.ts` covering empty / single / multiple accepts, `--chain` match / miss, and max-amount below / equal / above.

**5 — `stringToHex`.** Dropped the manual hex encoder in `signers/rpc.ts`; using `viem.stringToHex` directly.

**6 — `isAddress`.** Replaced the `ADDRESS` regex in `signers/rpc.ts` and `signers/module.ts` with `viem.isAddress(address, { strict: false })`. One source of truth, checksum-friendly.

**7 — Error taxonomy.** `SignerResolutionError` now extends `CliError` with exit code 6. Removed the special-case branch in `bin.ts` — one dispatch path.

**8 — Dead `SignerFactory` export.** Dropped from public exports. Will add back when a concrete consumer lands.

**9 — Privy/CDP docs.** Acknowledged as pre-publish: I'll run each example against the real SDK before cutting 0.2.0 to npm, and update the README if the actual shape differs.

**10 — `adaptAccount` cast.** Left as-is with the existing comment. Agreed this is an upstream concern for `@x402/evm`; will file a separate issue there if we hit it again.

Tests: 19 → 35 across four files (`chain.test.ts`, `signers.test.ts`, `helpers.test.ts`, `pay.test.ts`). Biome clean, typecheck clean, full `pnpm test` passes.

vraspar:
### SDK Review — Round 2

Found 0 new/unresolved issues since commit b2124e6 (axes: tests, conventions, dead code, SDK design).

Verified the round-2 commits (8bdc082, 2947c08) against the round-1 review:
- Checked out the branch, ran `pnpm install && pnpm test && pnpm typecheck` in `packages/cli` — **35/35 tests pass, typecheck clean**.
- Inspected each fix for behavior-preserving semantics and absence of regressions.

1. RPC signer BigInt JSON serialization (blocker) — replacer converts `bigint → decimal string` at `signers/rpc.ts:53-55`. New test at `tests/signers.test.ts:78-143` invokes `signTypedData` with real EIP-3009-shaped `value` / `validAfter` / `validBefore` BigInts and asserts the outgoing `eth_signTypedData_v4` body carries them as decimal strings — real regression guard, not just a shape check.
2. RPC signer test that never called `signTypedData` — covered by the same new test.
3. `pay()` untested — new `tests/pay.test.ts` covers 6 scenarios with stubbed `fetch`.
4. `pickAccept` / `enforceMaxAmount` untested — new `tests/helpers.test.ts` covers empty / single / multi accepts, `--chain` match/miss, and max-amount below/equal/above.
5. Manual hex encoder — replaced with viem's `stringToHex`.
6. ADDRESS regex duplication — replaced with `isAddress(addr, { strict: false })` in both `signers/rpc.ts` and `signers/module.ts`.
7. `SignerResolutionError` outside the `CliError` taxonomy — now extends `CliError(msg, 6)`; `bin.ts` collapsed to a single `instanceof CliError` dispatch.
8. Unused `SignerFactory` export — dropped from `types.ts` and the package root.

**Still open:**

- #9 (Privy / CDP README examples) — author correctly left as manual-verify-before-publish; requires installing the real provider SDKs. No code change needed in this repo.
- #10 (`adaptAccount` cast through `unknown`) — author correctly treated as an upstream concern for `@x402/evm`. Not this PR's problem.

**New:** none.

---

**Design decision (SDK/CLI boundary, from the follow-up comment):** author chose Option 2 — keep `pay()` in `@x402r/cli`, export it programmatically. Reasonable for a 1-verb HTTP-side surface; the migration to a separate `@x402r/pay` package stays mechanical if the follow-up verbs (`dispute`, `evidence`, `verify`) land later. `createPayerClient` naming collision remains something to settle before any future HTTP-payer client abstraction ships.

**One minor structural nit** (not blocking, not worth a fix on this PR): `pickAccept` and `enforceMaxAmount` are now `export`-ed from `pay.ts` purely to let the test import them. They are not re-exported from `src/index.ts` and `package.json` only declares `.` as an entry, so they are internal-by-convention but part of the source file's export shape. Moving them to `src/pay/policy.ts` would make the internal/public split structural. Mention for a future cleanup.

No further review rounds needed. PR looks ready.

---

vraspar:
pickAccept and enforceMaxAmount are now export-ed from pay.ts purely to let the test import them. They're not re-exported from src/index.ts, and package.json only declares the . entry, so they're internal-by-convention but technically part of the source file's public shape. Cleaner long-term would be moving them to src/pay/policy.ts so the internal/public split is structural. Not a blocker.

A1igator:
Took the structural nit too (fe0bb02). Moved `pickAccept` and `enforceMaxAmount` to `src/pay/policy.ts` and promoted `src/pay.ts` to `src/pay/index.ts`. The policy helpers are now reachable only via `./pay/policy.js`, which isn't re-exported from the package root, so the internal/public split is structural rather than convention.

Also tightened the vitest coverage exclude from `src/**/index.ts` to `src/index.ts` + `src/bin.ts` so the real logic in `src/pay/index.ts` (and incidentally `src/signers/index.ts`) stays measured.

35/35 tests still passing. Thanks for the thorough passes.

vraspar (approved):

PR #122, opened 2026-05-06
core+sdk: lift to authCapture contract surface (PR 1 of 4)

vraspar:
## Summary

PR 1 of 4 in the [authCapture SDK migration plan](https://github.com/BackTrackCo/x402r-notes/blob/main/plans/AUTHCAPTURE_SDK_MIGRATION.md). Lifts the SDK to track the merged contracts ABI from [BackTrackCo/x402r-contracts#34](https://github.com/BackTrackCo/x402r-contracts/pull/34) plus the canonical CREATE2 deploy at `df0d2af` (which renamed `PaymentIndexHook → PaymentIndexRecorderHook`). Subsequently rebound to the audited `commerce-payments@v1.0.0` deployment of `AuthCaptureEscrow` in `7ba81da`; `PaymentIndexRecorderHook` chain singleton now lives at `0x358ECA14fFD51e63D2Bb8DDE3aBAA14f8D5274C3` (v1.0.1 salt namespace).

Pure migration. No new capability — autoCapture, Permit2, and wire-format glue land in PRs 2/3/4.

Bumps to 0.3.0. Clean break, no shims.

## What's done in this PR

**Phase 1 — initial lift to authCapture surface**
- ABI regen from contracts main; old `*Recorder*` ABIs replaced by `*Hook*` ones
- Slot-getter selectors updated (`AUTHORIZE_PRE_ACTION_CONDITION`, etc.; `FEE_RECEIVER`)
- `paymentOperator.release()` / `refundInEscrow()` / `refundPostEscrow()` SDK actions renamed to `capture` / `voidPayment` / `refund` to match the contract methods; `void` is full-only and drops `amount` (no shim)
- Plugin terminology: `recorder→hook` across types, files, exports, factory functions, config fields (`recorders→hooks`, `RecorderSingletonAddresses→HookSingletonAddresses`, `recorderCombinator*→hookCombinator*`, `paymentIndexRecorderHook`, `authorizationTimeRecorderHook`, etc.)
- `ConditionConfig` → `PluginConfig` type alias
- Event renames in core + SDK watch path (`AuthorizeExecuted`, `CaptureExecuted`, `VoidExecuted`, `RefundExecuted`); `FeesDistributed.arbiterAmount→operatorAmount`
- Drop `usdcTvlLimit` (no longer in canonical deploys)
- Drop SKALE Base chain (Cancun-vs-Shanghai bytecode incompatibility)
- `actions/recorder/` directory → `actions/hook/` (preserved git history via `git mv`)
- pnpm 10 workspace fix: `@x402r/sdk` and `@x402r/helpers` now use `workspace:^` for `@x402r/core` so local edits resolve through the workspace
- File/function renames: `release.ts→capture.ts`, `refundInEscrow.ts→voidPayment.ts`, `refundPostEscrow.ts→refund.ts`, allowance helpers (`approvePostEscrowRefund→approveRefundAllowance`, `getPostEscrowRefundAllowance→getRefundAllowance`)
- `OperatorSlots` return-shape field renames (`authorizeRecorder→authorizeHook`, `feeRecipient→feeReceiver`, etc.)
- `options.feeRecipient → options.feeReceiver` on `MarketplaceOperatorOptions` / `DeliveryProtectionOperatorOptions` / `ArbiterSetupOptions`
- Final canonical CREATE2 addresses pinned in `packages/core/src/config/index.ts`. After the `7ba81da` rebind to canonical `commerce-payments@v1.0.0` `AuthCaptureEscrow`, `x402rChains` is `{8453: Base, 84532: Base Sepolia}` — the chains where the canonical escrow is currently deployed. Other EVMs return as canonical coverage extends upstream.

**Round 2 cleanup (post-review feedback from automated + manual passes)**
- Stale JSDoc + `'release'` op-name string sweep in `errors.test`/`error-wrapping.test` (`2bf416a`)
- `recorders → hooks` field rename on `HookCombinator` factory parameter types + `createRecorderProvider → createHookProvider` (`b20f69f`)
- `refundInEscrowConditionAddress → voidConditionAddress` on the 4 public preset types (`d16baf3`)
- `can-execute.test`, `deploy-fixtures.ts`, `vitest.config.ts` aligned to authCapture surface (`2e45496`)
- SDK README `.extend()` sample + 0.3.0 changeset operator-method section + core README export list rewrites (`c4491e6`)
- Examples: `release-escrow → capture-payment` rename + stale-terminology cleanup (`c61716e`)

**Round 2 correctness + new features**
- Cross-operator scoping on hook reads: optional `operatorAddress?: Address` filter on all 5 core hook actions (`6f285f5`); SDK `query.*` methods auto-scope to the configured operator by default (`e870e85`). Closes the silent-mingled-data vector on the chain-singleton `PaymentIndexRecorderHook`.
- Partial-capture migration story rewrite per teammate direction — replaces "capture-then-refund via `ReceiverRefundCollector`" with `capture(merchantAmount)` + `voidPayment(remainder)` (`4107ea8`). Two calls, no allowance, no collector setup.

**Round 2 fork-test omnibus**
- Filename renames: `authorize-release → authorize-capture`, `post-release-refund → post-capture-refund`, `query-recorder → query-hook` (`a5b03ea`)
- `partial-refund.fork.test` rewritten as a partial-capture regression guard (`697df91`)
- Identifier sweep across 9 fork tests + delete `usdcTvlLimit` test + obsolete Scenario 8c (`eb37dd9`)
- `FORK_BLOCK` dropped (public Base Sepolia RPC has narrow archive horizon — pinning fails) + knip ignores for autogen ABIs and the intentional `commercePaymentsAuthCaptureEscrow` alias + inlined `CreateHookProviderOptions` (`f810236`)
- `FREEZE_CONDITION/UNFREEZE_CONDITION → *_PRE_ACTION_CONDITION` rename in deploy.fork.test (`d4a9251`)

**Round 3 review**
- `getAddress()` normalize on the operator filter — fixes silent wrong-data when caller passes lowercased operatorAddress (`94d89a4`)
- `releaseConditionAddress → captureConditionAddress` on preset interfaces — parallel to the `refundInEscrow → void` rename, missed in Phase 1 (`29c0a33`)
- Pagination perf caveat in `createHookProvider` JSDoc + dispute-resolution example comment fix (`eefae04`)

**Round 4 review**
- CodeQL `let total = 0n` → `let total: bigint` (`6260012`)

**Architecture rebind + dep bump (post-Round 4)**
- Config rebound to canonical `commerce-payments@v1.0.0` `AuthCaptureEscrow` deployment (audited upstream); downstream escrow-bound contracts (PaymentOperatorFactory, EscrowPeriodFactory, FreezeFactory, RefundRequestFactory, ReceiverRefundCollector, PaymentIndexRecorderHook) move to the new `x402r-canonical-v1.0.1::*` salt namespace (`7ba81da`)
- `pnpm.overrides` adds `fast-uri >= 3.1.2` to clear two newly-published GHSAs (`da59746`)

**Round 5 review**
- `voidPayment.ts` JSDoc rewrite to reference partial-capture (was still pointing at the round-1 capture-then-refund-via-ReceiverRefundCollector story); reclaim recovery doc note added to changeset + SDK README; `getHookPaymentInfo` zero-address check uses `getAddress()` for pattern consistency (`4731761`)
- `partial-refund.fork.test` token-balance assertions added — proves tokens actually moved end-to-end (was state-only assertions before) (`a2582ee`)

## Migration: partial in-escrow refund (R-25)

The new `escrow.void()` is full-only — calling it empties the entire authorization in one transaction. Partial in-escrow refunds use **partial capture + void remainder**:

```ts
const { capturableAmount } = await client.payment.getAmounts(paymentInfo)
const merchantAmount = capturableAmount - refundToPayer

// 1. Capture only what the merchant keeps (decrements escrow.capturableAmount)
await client.payment.capture(paymentInfo, merchantAmount, '0x')

// 2. Void the remainder — returns the rest to the payer
await client.payment.voidPayment(paymentInfo)
```

Two calls, no allowance, no `ReceiverRefundCollector`. See [.changeset/sdk-authcapture-lift.md](.changeset/sdk-authcapture-lift.md) for the full migration note + the [SDK README .extend() example](packages/sdk/README.md#extending) for a dispute-flow integration. Post-capture refunds (after `capture()`) still use `payment.refund()` via `ReceiverRefundCollector` with a pre-staked allowance.

## Out of scope — next PRs (per the migration plan)

- `@x402r/helpers` scheme filter `'commerce' → 'authCapture'` → PR 2
- `autoCapture` plumbing → PR 2
- Permit2 support → PR 3
- `x402rDefaults()` helper + new wire-format type re-exports → PR 4
- Bumping `@x402r/evm` peerDep version → PR 2
- Hook query scoping JSDoc + opt-out work for direct `@x402r/core` consumers → PR 4 per the updated migration plan in `x402r-notes`

## Out of scope — separate follow-up issues

- `FeeAddresses.operatorFeeRecipient` reads `FEE_RECEIVER` slot — wire/onchain naming drift, design decision deferred (rename for consistency vs. keep as a consumer-friendly alias)
- Cursor-based pagination on `PaymentIndexRecorderHook` — needs a contract change; current SDK pagination is correctness-correct but linear in `total` under heavy cross-operator mingling

## Test plan

- [x] `pnpm check` (biome) clean
- [x] `pnpm typecheck` clean across all packages
- [x] `pnpm build` clean across all packages
- [x] `pnpm --filter @x402r/core test:unit` — **211 passed**
- [x] `pnpm --filter @x402r/sdk test` — **196 passed**
- [x] `pnpm --filter @x402r/core test:fork` (CI) — Fork Tests check green
- [x] `pnpm knip` (Publish Validation) — green
- [x] All 9 CI checks green at HEAD


A1igator:
### SDK Review

Found 9 issues (axes: tests, conventions, dead code, example alignment):

1. **[Example alignment]** sdk README ships broken public API names. The `.extend()` code sample on L74 calls `client.payment.refundInEscrow(paymentInfo, refundableAmount)` — wrong method name and wrong arity (the new `voidPayment` takes only `paymentInfo` per L13 of the changeset).

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/sdk/README.md#L43-L106

2. **[Example alignment]** Changeset for 0.3.0 contradicts what shipped on this branch. L12 says SDK `release` is "still exported" (renamed in 746e2af); L14 says SDK `refundPostEscrow` is "unchanged for now" (renamed in a77cb71); L15 says `OperatorSlots.feeRecipient` is "kept for now" (renamed in cd7ed3f).

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/.changeset/sdk-authcapture-lift.md#L12-L15

3. **[Conventions]** `recorders` field leaks on public Hook factory parameter types. `ComputeHookCombinatorAddressParameters.recorders` and `DeployHookCombinatorParameters.recorders` retain the old name after the changeset claims `recorder→hook` was applied everywhere; the on-chain ABI param is `_hooks`.

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/core/src/deploy/factories.ts#L327-L356

4. **[Conventions]** `createRecorderProvider` and `name: 'recorder'` ship as public surface. The function is re-exported through `query.ts:5,21`; the literal `'recorder'` string surfaces in resolver telemetry/logs and resolution traces.

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/sdk/src/resolver/providers.ts#L23-L32

5. **[Conventions]** `refundInEscrowConditionAddress` field across 4 public preset types. The contract method is `void` and the SDK function is `voidPayment`; the field appears 27 times across `MarketplaceOperatorPreview`, `MarketplaceOperatorDeployment`, `DeliveryProtectionOperatorPreview`, and `DeliveryProtectionOperatorDeployment` exported types.

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/core/src/deploy/presets.ts#L233-L244

6. **[Tests]** `can-execute.test.ts` passes invalid `'RELEASE_CONDITION'` literal at five call sites. Not a member of the new `ConditionSlot` union (`AUTHORIZE_PRE_ACTION_CONDITION` / `CAPTURE_PRE_ACTION_CONDITION` / etc.); the test only passes because `getConditionAddress` is mocked and the test file is excluded from typecheck. File was untouched by this PR's rename sweep.

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/sdk/tests/can-execute.test.ts#L39-L113

7. **[Tests]** `deploy-fixtures.ts` rename is partial. Commit c3b49ba6 only renamed `feeRecipient`→`feeReceiver`; left `releaseCondition`/`releaseRecorder`/`refundInEscrowCondition`/`refundInEscrowRecorder`/`refundPostEscrowCondition`/`refundPostEscrowRecorder` field names. These will fail at runtime against the new ABI when fork tests are re-enabled.

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/core/tests/setup/deploy-fixtures.ts#L129-L134

8. **[Dead code]** Coverage excludes reference deleted files. `release.ts`, `refundInEscrow.ts`, `refundPostEscrow.ts`, `approvePostEscrowRefund.ts`, `getPostEscrowRefundAllowance.ts` all gone; their successors (`capture.ts`, `voidPayment.ts`, `refund.ts`, `approveRefundAllowance.ts`, `getRefundAllowance.ts`) silently start counting toward coverage instead of being excluded as thin write forwarders.

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/core/vitest.config.ts#L17-L31

9. **[Example alignment]** core/README.md lists `ConditionConfig` as an exported type. Renamed to `PluginConfig` in commit 9de5694; the README now documents an export that no longer exists.

   https://github.com/BackTrackCo/x402r-sdk/blob/c61716efcd6006c662a3d8bb9cc460460a50e52d/packages/core/README.md#L20

---

vraspar:
Did a four-theme review pass (ABI + addresses, action renames, terminology, tests + examples). The public-API surface is clean — ABIs match contracts at `df0d2af`, all 20 pinned addresses verified by independent CREATE2 derivation, renamed actions wire to the right contract functions, `void` correctly drops `amount` (no shim), 198 core unit tests + 192 SDK unit tests green, examples typecheck at parity with main.

Below the public-API surface, several things are half-migrated. All 13 fork tests fail (0 passing). Mostly traces back to a single fixture file that was only partially updated.

## Blockers

1. **`packages/core/tests/setup/deploy-fixtures.ts`** only renamed `feeRecipient → feeReceiver`. Every other slot still uses old names (`authorizeRecorder`, `releaseCondition`, `refundInEscrowCondition`, etc.). The new `PluginConfig` shape rejects this, breaking every fork test that uses the fixture.
2. **`packages/core/tests/integration/config-addresses.fork.test.ts:8`** imports nonexistent `usdcTvlLimitAbi`, tests dropped `config.usdcTvlLimit` field, and only iterates Base Sepolia rather than all 12 chains.
3. **`packages/core/tests/integration/delivery-protection.fork.test.ts`** calls deleted methods `payment.release()` (lines 212, 225) and `payment.refundInEscrow()` (lines 262, 298, 337); reads non-existent slots (lines 119, 125, 131); has stale `feeRecipient` (line 62).
4. **`packages/core/tests/integration/approve-refund-request.fork.test.ts`** calls `merchant.payment.refundInEscrow(paymentInfo, AMOUNT)` (lines 132, 160, 237, 320) and `merchant.payment.release(...)` — neither method exists.
5. **`packages/core/tests/integration/deploy.fork.test.ts:27,215`** passes `feeRecipient` to `MarketplaceOperatorOptions`; the deploy presets type rejects this.
6. **`packages/core/tests/integration/query-recorder.fork.test.ts:91-104`** uses old `OperatorConfig` struct shape (`authorizeRecorder`, `releaseCondition`, `refundPostEscrowCondition`); ABI encoding will fail at runtime.
7. **`packages/core/vitest.config.ts:17,26-28,31`** coverage exclude list points at deleted files (`release.ts`, `refundInEscrow.ts`, `approvePostEscrowRefund.ts`, `refundPostEscrow.ts`, `getPostEscrowRefundAllowance.ts`).
8. **`examples/shared/anvil-setup.ts:71`** and **`packages/core/tests/setup/anvil.ts:111`** — the `FORK_BLOCK` predates the canonical CREATE2 deployments. Fork-based examples and tests can't run end-to-end without a fresher block.
9. **`packages/sdk/README.md`** still lists/uses old method names: line 43 (action group list), line 74 (`extend()` example calls `refundInEscrow`), lines 98, 102, 106 (prose in extension narrative). User-facing.
10. **`.changeset/sdk-authcapture-lift.md`** doesn't document the rename cascade. Currently claims SDK-side `release` still exists (line 12), `refundPostEscrow()` is "unchanged for now" (line 14), `refundInEscrow` still exists (line 13), `OperatorSlots` keeps `feeRecipient` (line 15). Anyone reading the changeset to migrate will miss half the breaking changes.

## Should-fix

- `packages/core/src/deploy/factories.ts:327,338,344,356` — `ComputeHookCombinatorAddressParameters.recorders` and `DeployHookCombinatorParameters.recorders` should be `hooks` to match the rest of the rename. Callsite in `presets.ts:1110` follows.
- `packages/sdk/src/resolver/providers.ts:23,26,32` — `createRecorderProvider` and `name: 'recorder'` should be `createHookProvider` / `'hook'`. Consumers in `sdk/src/actions/query.ts:5,21`, `createResolver.ts:7`, and tests follow.
- `packages/core/src/deploy/presets.ts` — internal symbols `refundInEscrowConditionAddress` / `refundInEscrowLegs` (~17 occurrences). Public API unaffected; consistency for maintainers.
- `packages/sdk/tests/can-execute.test.ts` passes `'RELEASE_CONDITION'` 5x; the new `ConditionSlot` enum has only `AUTHORIZE_/CHARGE_/CAPTURE_/VOID_/REFUND_PRE_ACTION_CONDITION`. Passes at runtime only because mocks bypass the slot read.
- `packages/core/tests/integration/authorize-release.fork.test.ts:86,146,147` reads `config.feeRecipient` (doesn't exist on `OperatorSlots`); also has slot-name drift.
- `tsconfig.build.json` excludes `tests/`, and `tsconfig.json` extends without overriding. That's why all this fork-test drift compiles silently. Consider including `tests/` in typecheck.

## Nice-to-have

- PR description nits: claims to track contracts at `9786579` but the address pinning (specifically `PaymentIndexRecorderHook` at `0x16CF…084E`) reflects later contracts state at `df0d2af`; claims `void` has a back-compat shim that ignores `amount` — there isn't one (the code is correctly clean-break, matching the plan).
- Stale JSDoc/comments: `actions/operator/authorize.ts:19` ("Use `release()`"), `deploy/presets.ts:362,419,1359,1061,1106,1158`, `actions/index.ts:101` (`// Recorder`), `sdk/src/types.ts:125`.
- `'release'` string literals in `tests/errors.test.ts:64,65,77` and `tests/error-wrapping.test.ts:24,43,47,54,61,68,103,107,123,134` — cosmetic.
- `packages/core/src/actions/fees/types.ts:17-18` exposes `operatorFeeRecipient` field while reading the `FEE_RECEIVER` slot — wire/onchain naming inconsistency. Decide intent (consumer-friendly alias vs. align to `operatorFeeReceiver`).

## On scope

The fork-test breakage is the dominant issue. Most blockers (1–6) collapse to "update `deploy-fixtures.ts` to the new `PluginConfig` shape, then sweep the four fork tests that consume it." Worth doing in this PR rather than letting the follow-up PRs in the migration plan inherit the debt.

vraspar:
Sharing some design context that came up in review — not a blocker, useful framing for the broader team.

## How the refund flow changes for users

**Old (`refundInEscrow`):** authorize → escrow holds funds → one call to refund a partial amount. No allowance, no merchant capital required.

**New:**
- *Full pre-capture refund:* `void(paymentInfo)` — clean.
- *Partial pre-capture refund (the old common case):*
  1. `capture(paymentInfo)` — funds move from escrow to merchant wallet
  2. `ERC-20.approve(ReceiverRefundCollector, amount)` if not pre-approved
  3. `refund(paymentInfo, amount, ReceiverRefundCollector)` — collector pulls from merchant wallet, sends to payer

Merchants are most impacted: capital efficiency hit (refund liquidity must live in their wallet, not escrow), new allowance management surface, and a trust shift — the protocol no longer guarantees refundability after capture.

## How Stripe handles this

Stripe is the closest mental analogue but the comparison breaks where it matters most. Stripe holds funds across the full lifecycle. `POST /v1/refunds` with optional `amount` works the same whether the charge is captured or not — pre-capture acts as a void, post-capture pulls from Stripe's balance owed to the merchant. **The merchant never pre-approves anything.** Stripe is the custodian.

The new x402r design echoes Stripe's *conceptual* void/refund split, but after capture funds leave protocol custody. The Stripe-style "platform holds funds" model isn't achievable on-chain without trusted custody, which defeats the protocol's purpose.

## Alternative on-chain designs worth knowing

| Design | Trade-off |
|---|---|
| **Partial-void in escrow** (the old way) | Requires more contract state; supports partial refund without leaving escrow |
| **Partial capture** (`capture(amount)`, remainder auto-voided) | Same UX as old `refundInEscrow` but cleaner contract semantics — escrow keeps custody until explicitly captured |
| **Merchant refund-pool deposit** | Merchant pre-funds a protocol-held balance for refunds; like a security deposit |
| **Refund-claim ERC-20 issuance** | Payer gets a token redeemable against merchant pool; overkill for most cases |

## My take

The new design optimizes for **contract simplicity** at the cost of **merchant complexity**. That's defensible for v1 — the contract surface is cleaner and refunds can happen indefinitely after capture without escrow retaining state.

If I were designing fresh, **partial capture** is probably the cleanest spot in the design space: keeps custody, supports partial refund implicitly, no allowance gymnastics, no capital lockup for merchants. Worth considering for a v2 if the merchant friction shows up in practice.

For now: ship the current design, but track "merchant capital lockup" and "allowance misconfiguration" as real product friction worth measuring.

vraspar:
One more correctness finding from a focused plugin-layer review — splitting from the prior cleanup comment because this is about query semantics, not test scaffolding.

## Cross-operator data mingling on `PaymentIndexRecorderHook` reads

The canonical `PaymentIndexRecorderHook` singleton at `0x16CF…084E` aggregates `payerPaymentCount` across every operator that routes through it. The contract docstring at `PaymentIndexRecorderHook.sol:22-29` explicitly says per-operator views require reading `paymentInfo.operator` from the returned struct and filtering caller-side.

The SDK doesn't propagate that. None of these surface the aggregation, default-filter, or expose a scoping option:

- `packages/core/src/actions/hook/getPayerPayment.ts`
- `packages/core/src/actions/hook/getPayerPayments.ts`
- `packages/core/src/actions/hook/getReceiverPayment.ts`
- `packages/core/src/actions/hook/getReceiverPayments.ts`
- `packages/core/src/actions/hook/getHookPaymentInfo.ts`
- `packages/sdk/src/resolver/providers.ts:26-76` (`createRecorderProvider`)
- `packages/sdk/src/actions/query.ts` (`getPayerPayments(payer)`)
- `packages/sdk/src/types.ts:177-181` (`QueryActions` interface — no operator-scoping option)

`createEventProvider` is correctly scoped per-operator via `operatorAddress`. The recorder provider, which fires before events in the resolver chain, isn't.

**Why it matters:** any SaaS running two operators on the same chain (e.g. marketplace fees + direct sales) gets mingled query results without realizing it. This isn't a test-scaffold issue — it's wrong data being returned to integrators by default.

**Suggested fix:** add an optional `operatorAddress?: Address` param on each read function that, when set, filters by `info.operator`. JSDoc each read path with a clear warning. The fancier "auto-filter by configured preset" can be a follow-up.

Doesn't need to land in this PR if you want to keep it focused on the migration — but worth getting in before downstream consumers wire up multi-operator setups against the SDK as-is.

vraspar:
Tier-2 verification pass on the latest commits — all prior blockers resolved (`deploy-fixtures.ts` is a semantic match, not just textual rename; fork tests demonstrably pass end-to-end; cross-operator scoping defaults to the configured operator at the SDK layer with an explicit opt-out via `@x402r/core/actions/hook/*` documented in JSDoc, which is stronger than the opt-in I'd originally suggested). 487/487 tests green; `pnpm typecheck` and `pnpm build` clean.

Three small follow-ups worth folding in before merge:

1. **Address case-sensitivity in the new operator filter.** `packages/core/src/actions/hook/getPayerPayment.ts:39`, `getPayerPayments.ts:50`, `getReceiverPayment.ts:39`, `getReceiverPayments.ts:50`, `getHookPaymentInfo.ts:40` — compare `p.operator === operatorAddress` strictly. Chain returns checksummed; a lowercased caller-supplied address silently returns empty. Normalize via `getAddress()` on both sides before comparison. Add a test that passes a lowercased `operatorAddress` and asserts non-empty results so this doesn't regress.

2. **Pagination perf cliff under heavy cross-operator mingling.** `packages/sdk/src/resolver/providers.ts:46-79` — `createHookProvider` loops `offset += pageSize` until `offset < total`, where `total` is on-chain pre-filter. With many co-tenants on the canonical singleton this can issue many RPC roundtrips and toss most results. Not a correctness bug, just a perf surprise. JSDoc note now; cursor-based pagination is the real fix later.

3. **`releaseConditionAddress` naming drift.** `packages/core/src/deploy/presets.ts:1024, 1035` — public field on `DeliveryProtectionOperatorPreview` / `Deployment` is still `releaseConditionAddress`. Commit `d16baf3` renamed `refundInEscrowConditionAddress → voidConditionAddress` but didn't touch this one. The action method was renamed `release → capture` everywhere else, so this is the matching rename — should be `captureConditionAddress` for consistency.

After these, the PR is ready to land as the migration foundation.

A1igator:
### SDK Review — Round 2

Found 2 new issues since `c61716ef` (axes: conventions, example alignment):

**Resolved since last round:** SDK README `.extend()` sample method/arity; changeset claims vs shipped renames; `recorders` field on HookCombinator factory params; `createRecorderProvider`/`name: 'recorder'`; `refundInEscrowConditionAddress` field on preset types; `RELEASE_CONDITION` literal in can-execute.test; deploy-fixtures partial rename; vitest coverage exclusion paths; core/README `ConditionConfig` listing.

**Still open:** none.

**New:**

1. **[Conventions]** `releaseConditionAddress` survives the `release→capture` rename on public preset types. The changeset states "SDK function and on-chain method both renamed; no back-compat alias", but `DeliveryProtectionOperatorPreview` and `DeliveryProtectionOperatorDeployment` still expose a field named `releaseConditionAddress` — same pattern as the now-resolved `refundInEscrowConditionAddress`. The address is wired into `capturePreActionCondition` per L1120; the slot, the SDK action (`payment.capture`), and the on-chain method are all `capture`. Local var (L367, L1088), public fields, and 6 in-tree test references (`presets.test.ts` L767/L859/L900, `deploy.fork.test.ts` L242/L277, `delivery-protection.fork.test.ts` L120) all carry the old name.

   https://github.com/BackTrackCo/x402r-sdk/blob/d4a9251193b22ad30a02aaa52f88abde2c1ab4c6/packages/core/src/deploy/presets.ts#L1020-L1043

2. **[Example alignment]** Stale partial-in-escrow refund migration in dispute-resolution example. The comment block says `partial in-escrow refunds use capture-then-refund via ReceiverRefundCollector` — that path was the migration story in the original changeset, but commit `4107ea8d` rewrote the canonical migration to **partial-capture + void remainder** (no allowance, no ReceiverRefundCollector). The example was missed in the rewrite sweep, so it now contradicts both the changeset migration note (L55-L75) and the SDK README L83-L92. ReceiverRefundCollector is post-capture-only on the new surface.

   https://github.com/BackTrackCo/x402r-sdk/blob/d4a9251193b22ad30a02aaa52f88abde2c1ab4c6/examples/scenarios/dispute-resolution.ts#L135-L137

---

vraspar:
## Round 2 triage + fixes

Quick summary of what landed in `94d89a4` / `29c0a33` / `eefae04` and what I deliberately didn't touch.

### Addressed (3 commits)

**`94d89a4` — operatorAddress filter case-sensitivity** (the only correctness item)

vraspar's Tier-2 finding was a real wrong-data vector: the W2 hook filter used strict `===` against the operator address, so a caller passing the lowercased form (URL params, env vars) would silently miss matches against the chain's checksummed return value. Wraps both sides of every operator compare in viem's `getAddress()` across all 5 hook reads (`getHookPaymentInfo`, `getPayer/ReceiverPayment`, `getPayer/ReceiverPaymentsFromHook`). Adds 5 regression tests in `hook.test.ts` that derive a checksummed mock-return from the lowercased fixture via `getAddress()` and assert the filter still matches — without the fix these tests silently return empty/null even though the addresses point at the same operator.

**`29c0a33` — `releaseConditionAddress` → `captureConditionAddress` rename**

vraspar Tier-2 + A1igator round 2 both flagged this independently. Same pattern as the Phase 1 `refundInEscrowConditionAddress → voidConditionAddress` rename — applied to the parallel field that was missed in the same scan. Touches `presets.ts` public interfaces (DeliveryProtectionOperator{Preview,Deployment}) + 12 internal cascade refs + 3 test files. The MarketplaceOperator interfaces already use `voidConditionAddress` only and never had a `releaseConditionAddress` field, so no edits there.

**`eefae04` — JSDoc perf caveat + dispute-resolution example fix**

Two small docs items.

- vraspar Tier-2: extends the `createHookProvider` JSDoc with a perf caveat for the cross-operator pagination case (loop iterates over the unfiltered `total` and discards most pages client-side under heavy mingling). Self-flagged as JSDoc-now / cursor-based-pagination-later; agree.
- A1igator round 2: `examples/scenarios/dispute-resolution.ts:135-137` still recommended "capture-then-refund via `ReceiverRefundCollector`" — contradicts the partial-capture rewrite that landed in `4107ea8` (changeset migration note + SDK README). Rewrites the comment to point at the partial-capture pattern and references the canonical sources of truth.

### Intentionally not touched

**CodeQL inline lints (`let total = 0n`, 4 duplicates)** — the `0n` initializer is deliberate for TypeScript's definite-assignment analysis and the 4 alerts are 1 issue × 2 functions × 2 commit re-runs. Removing the initializer and relying on TS narrowing through the do-while body works at the moment but adds zero defensive value and is fragile to future loop refactors. Recommend dismissing in the CodeQL UI as `won't fix / intended`.

**`FeeAddresses.operatorFeeRecipient` reads `FEE_RECEIVER` slot** — wire/onchain naming drift flagged in vraspar's first audit round; design decision (rename for consistency vs keep as a consumer-friendly alias) deferred to a follow-up issue rather than mixed into this PR.

**`tsconfig` typecheck scope** — first-round audit suggested test-file drift compiles silently; investigation against the current configs showed `tsconfig.json` already includes `tests/**/*.ts`, so the original concern doesn't reproduce. No action.

**Cursor-based pagination on `PaymentIndexRecorderHook`** — the structural fix for the perf caveat. Genuinely follow-up work; needs a contract change.

CI watch starting now.

vraspar:
PR description checklist for the historical record before merge — none of these change code, just documentation drift accumulated over three review rounds. Worth folding into a fresh PR body so the merge commit reflects what actually shipped.

**Tracking claim** (summary, line 1)
> Lifts the SDK to track the merged contracts ABI from x402r-contracts#34 (HEAD `9786579`)

Actually tracks `df0d2af`. The pinned `PaymentIndexRecorderHook` (`0x16CF…084E`) and the `PaymentIndexHook → PaymentIndexRecorderHook` rename came from a follow-up contracts PR, not #34. Suggest rephrasing to point at the canonical deploy SHA.

**`void` back-compat shim claim** (under "What's done", bullet 3)
> the SDK function still accepts the parameter for back-compat but ignores it

No shim exists. `packages/core/src/actions/refund-budget/voidPayment.ts:9-13` only accepts `(operatorAddress, paymentInfo, data?)`. Clean break, consistent with the plan's "no shims" stance. Drop the parenthetical.

**Test count** (under "What's done")
> 198 unit tests green

Significantly stale — the latest core run shows **296 tests in 31 files** alone, plus SDK and a now-passing fork suite.

**"Out of scope — follow-up commits inside this PR (or fast-follows)" section**
Everything listed here actually landed in this PR:
- File renames (`746e2af`, `bc7a1ac`, `a77cb71`)
- `OperatorSlots` field renames (`cd7ed3f`)
- `options.feeRecipient → options.feeReceiver` (`c3b49ba`)
- Examples + integration fork tests (`c61716e`, `eb37dd9`, `a5b03ea`, `5ca3e5b`, `697df91`)
- Final canonical CREATE2 addresses (`c4467d8`, `1363c52`, `b20f69f`, `aecf9b0`)

Move these to "What's done" or delete the section.

**Migration paragraph (R-25)**
> Replacement: capture-then-refund via `ReceiverRefundCollector`

Superseded by partial-capture + void in `4107ea8`. The canonical migration story now lives in the changeset + SDK README. Either replace this paragraph inline with the partial-capture pattern, or point at those sources of truth.

**Worth adding to "What's done"** — rounds 2-3 wins not yet reflected in the description:
- Cross-operator scoping on hook reads — core opt-in filter (`6f285f5`) + SDK default-on scoping (`e870e85`)
- Address case-sensitivity normalization in the operator filter (`94d89a4`)
- `releaseConditionAddress → captureConditionAddress` parallel to the `refundInEscrow → void` rename (`29c0a33`)
- Partial-capture migration story rewrite (`4107ea8`)
- Fork-test omnibus sweep including `usdcTvlLimit` cleanup (`eb37dd9`)

**"Out of scope — next PRs"** section is still accurate. Optional: note that PR 4's scope now also includes the hook query scoping JSDoc + opt-out work per the updated migration plan in x402r-notes.

vraspar:
Round 4 — both items addressed.

**CodeQL `let total = 0n` (`6260012`)** — verified locally that TS strict accepts `let total: bigint` here via definite-assignment analysis through the do-while body, so the bot's first suggestion (declare without initializer) was the right call. Two-line fix in `getByPayer` and `getByReceiver`. Resolves all 6 inline alerts (1 issue × 2 functions × 3 commit re-runs).

**PR description rewrite** — applied your full checklist via GraphQL mutation (`gh pr edit` was silently failing due to the projects-classic deprecation):
- Tracking claim corrected: `9786579` → `df0d2af` (canonical deploy SHA)
- Dropped the false `void` back-compat shim parenthetical (verified `voidPayment.ts:9-13` accepts only `(operatorAddress, paymentInfo, data?)`)
- Test counts updated to current state: **211 core unit + 196 sdk + Fork Tests green in CI**
- Migration paragraph (R-25) rewritten inline with the partial-capture pattern, references the changeset + SDK README as sources of truth
- "Out of scope — follow-up commits inside this PR" section deleted; everything that lived there now appears in "What's done" with commit SHAs (split into Phase 1, Round 2 cleanup/correctness/fork-test, Round 3)
- Round 2-3 wins added to "What's done": cross-operator scoping, case-sensitivity normalize, `releaseConditionAddress → captureConditionAddress`, partial-capture migration rewrite, fork-test omnibus, JSDoc/release sweep, FORK_BLOCK fixes
- "Out of scope — separate follow-up issues" section added (`operatorFeeRecipient` naming + cursor-based pagination)
- "Out of scope — next PRs" section retained + noted PR 4 hook query scoping JSDoc work per the updated migration plan

A1igator:
### SDK Review — Round 3

Found 3 new/unresolved issues since `eefae04` (axes: conventions, tests, example alignment):

**Resolved since last round:** `releaseConditionAddress` survival on preset interfaces; stale partial-in-escrow refund migration in `dispute-resolution.ts`.

**Still open:** none.

**New:**

1. **[Conventions]** `29c0a33` `releaseConditionAddress → captureConditionAddress` rename left three internal `releaseCondition` references in the existence-tracking pipeline. The address variable was renamed everywhere, but the entry name registration on L1207 (`name: 'releaseCondition'`), the `exists` field + map-key string literal on L1271 (`releaseCondition: existsMap.get('releaseCondition') ?? false`), and the consumer site on L1356 (`exists.releaseCondition`, with the comment two lines above reading "OrCondition for capture") all still carry the old name. Same partial-sweep pattern Round 1 caught on `refundInEscrowConditionAddress` and Round 2 caught on the `Address`-suffixed field.

   https://github.com/BackTrackCo/x402r-sdk/blob/da59746b8513123898015d07091efb5d06ee71b8/packages/core/src/deploy/presets.ts#L1206-L1214

2. **[Tests]** Two stale `releaseCondition` test descriptions in the marketplace `previewMarketplaceOperator` block. The `it(...)` titles on L70 ("releaseCondition equals escrowPeriodAddress when freeze disabled") and L101 ("releaseCondition is AndCondition when freeze enabled") describe pre-rename behavior; the bodies assert `result.operatorConfig.capturePreActionCondition` (L94). `29c0a33` updated three assertions in the delivery-protection describe block but skipped these two marketplace describe descriptions. Reading the test report after the rename, the descriptions name a slot that no longer exists.

   https://github.com/BackTrackCo/x402r-sdk/blob/da59746b8513123898015d07091efb5d06ee71b8/packages/core/tests/deploy/presets.test.ts#L70-L101

3. **[Example alignment]** Workspace `README.md` "Supported chains" section diverges from `7ba81da`. L44 lists 11 chains (Ethereum, Polygon, Arbitrum One, Optimism, Celo, Avalanche C-Chain, Monad, Linea, Ethereum Sepolia, plus Base + Base Sepolia); `7ba81da` shrinks `x402rChains` to `{84532, 8453}` only and the changeset states "All chains except Base mainnet (8453) and Base Sepolia (84532) dropped from x402rChains". Same drift class as Round 1 issue #2 — code coverage reduced, parallel public-facing doc not updated. (The "CREATE3 addresses" wording on L46 predates this PR; the chain-list mismatch is what this PR widens from 11-of-12 to 11-of-2.)

   https://github.com/BackTrackCo/x402r-sdk/blob/da59746b8513123898015d07091efb5d06ee71b8/README.md#L42-L46

---

vraspar:
Reviewed against `da59746` over two passes — operatorAddress filter (deep) and partial-capture migration (deep). The rename axis was validated mechanically in a separate sweep and is clean.

The genuinely new code (operatorAddress auto-scoping) is solid. The semantics change (partial-capture replacing partial-void) has a stale-doc blocker and a test gap on the only behavior worth regression-protecting in this PR.

### Asks

**1. `packages/core/src/actions/refund-budget/voidPayment.ts:16-19` — JSDoc contradicts the rewritten migration story.** Reads "For partial refunds, use capture-then-refund via ReceiverRefundCollector instead." That was the *previous* (round-1) replacement story; the changeset was rewritten in `4107ea8` to be capture-then-void. Anyone reading source instead of the changeset gets pointed at the wrong primitive (and at a flow that requires a pre-staked allowance, which the new flow does not). Replace with: "For partial in-escrow refunds, use partial capture: `capture(merchantAmount)` then `voidPayment()` to return the remainder to payer. Post-capture refunds (after `capture()`) use `refund()` with a pre-staked `ReceiverRefundCollector` allowance."

**2. `packages/core/tests/integration/partial-refund.fork.test.ts:65-110` — no token-balance assertions on the only behavior change in this PR.** Test only asserts `capturableAmount` deltas (`DEFAULT_AMOUNT → REFUND_AMOUNT → 0`). Doesn't verify that the merchant actually received `MERCHANT_AMOUNT` USDC after `capture()` or that the payer actually received `REFUND_AMOUNT` USDC after `voidPayment()`. The pattern exists 60 lines away in `approve-refund-request.fork.test.ts:167-174` — pre/post `balanceOf()` snapshots with delta assertion. State assertions on `capturableAmount` alone don't prove tokens moved, and this is the *only* fork test guarding the new partial-capture story end-to-end.

**3. PR description is stale post-`7ba81da`.** Two claims no longer hold: (a) "Final canonical CREATE2 addresses pinned … for all 12 chains in `x402rChains`" — `x402rChains` is now `{8453, 84532}`. (b) The reference to `PaymentIndexRecorderHook 0x16CF…084E` — singleton moved to `0x358E…74C3` at v1.0.1. Changeset (line 43, 48) is current; only the PR body needs an edit so reviewers don't compare against the wrong addresses.

**4. Reclaim recovery path for half-completed partial-capture flow is undocumented.** New flow is two txs (`capture` → `voidPayment`). If a merchant captures but never voids — crash, gas exhaustion, key loss between txs — the payer's remainder sits in escrow in `Authorized` state. Recovery exists at the contract layer (`reclaim` at `abis/generated.ts:408`, `PaymentReclaimed` event at line 642), but there's no `payment.reclaim()` SDK action and no mention in `.changeset/sdk-authcapture-lift.md` or `packages/sdk/README.md`. Either ship a one-line `payment.reclaim()` wrapper now, or add a "recovery if void doesn't land" note in the migration section pointing at the on-chain method. Without either, merchants reading SDK docs see a 2-tx flow with no documented failure mode.

### Small things (single sweep)

- `packages/core/src/actions/hook/getHookPaymentInfo.ts:36` — zero-address sentinel check (`info.operator === zeroAddress`) is the only spot in the 5 hook actions not using `getAddress()` normalization. Practically safe (zero is identity-cased), but inconsistent with the pattern the round-3 fix (`94d89a44`) established. Cosmetic.
- No fork test exercises multi-operator scoping on the canonical singleton — the very vector the new `operatorAddress?` filter fixes. Filter logic is well unit-tested (`hook.test.ts:344-432`), but a fork-level proof requires deploying two operators sharing one `PaymentIndexRecorderHook`. Track as backlog if rig setup is non-trivial; flag here so it doesn't get lost.
- `partial-refund.fork.test.ts` doesn't cover edge cases: `merchantAmount === 0` (full void via trivial capture), `merchantAmount === capturableAmount` (does `voidPayment()` revert when there's nothing to void?), and double-application. Low severity, fold into the convenience-helper PR.
- Convenience helper that bundles `getAmounts → capture → void` is intentionally deferred per the migration plan backlog — agree with the deferral, but worth confirming explicitly that the raw 2-call pattern is the intended customer-facing surface for v0.3.0.

The auto-scope change is end-to-end clean: filter logic correct on both sides of the comparison, `getAddress()` normalization applied (commit `94d89a44`), pagination advances by requested count not post-filter length (`providers.test.ts:216-237`), `getPayerPayment`/`getReceiverPayment` correctly narrowed to `| null` for operator-mismatch detection, SDK auto-scopes by default in `query.ts:26-30`, and direct-`@x402r/core` callers correctly preserve unfiltered behavior unless they opt in. Ship the asks above and this is good to merge from a correctness standpoint.

vraspar:
Round 5 — addressed.

**Asks #1 (`voidPayment.ts` JSDoc) — `4731761`** — rewrote the JSDoc to point at the partial-capture flow (matches changeset/README/example). The old text was a real miss in my W3 sweep; I'd updated `dispute-resolution.ts:135` and the README + changeset, but the source-level JSDoc on `voidPayment` slipped through.

**Asks #2 (balance assertions in `partial-refund.fork.test`) — `a2582ee`** — added `balanceOf` snapshots + delta assertions for receiver post-capture and payer post-void. Mirrors the pattern in `approve-refund-request.fork.test.ts:167-174`. Receiver delta accounts for the operator fee (`MERCHANT_NET = MERCHANT_AMOUNT * (10_000 - FEE_BPS) / 10_000`); payer delta is the full `REFUND_AMOUNT` since `void` is full-only and doesn't take fees. `FEE_BPS = 50n` hardcoded with reference to the source-of-truth in `deploy-fixtures.ts:47` — if it ever changes, this test fails loudly rather than silently passing.

**Asks #3 (PR body stale post-`7ba81da`)** — done. The "12 chains" claim and the `0x16CF…084E` reference are both updated in the PR body. New body now reflects `x402rChains = {8453, 84532}` per the canonical-escrow chain coverage and references the new `0x358ECA14…74C3` singleton. Round 4 + Architecture rebind + Round 5 sections added to "What's done".

**Asks #4 (reclaim recovery) — `4731761`, doc-only path** — added a "Recovery if `voidPayment` doesn't land" paragraph to both the changeset migration block and the SDK README dispute-flow section. Points at `AuthCaptureEscrow.reclaim(paymentInfo)` callable by the payer after `paymentInfo.refundExpiry`, with the raw `walletClient.writeContract` invocation since there's no SDK wrapper yet. **`payment.reclaim()` typed wrapper deferred to PR 2/4** per your own recommendation — agree the wrapper is fair migration-plan follow-up rather than scope creep on this PR.

**Small thing #1 (`getHookPaymentInfo.ts:36` zero-address check) — `4731761`** — wrapped in `getAddress()` for pattern consistency with the rest of the round-3 fix. Behaviorally a no-op (zero is identity-cased) but removes the only outlier in the 5 hook actions.

**Small things #2, #3 (multi-operator fork test rig + `partial-refund` edge cases) — backlog confirmed.** Both are real gaps but the rig work is non-trivial (deploying two operators sharing one `PaymentIndexRecorderHook`); explicit follow-up items.

**Small thing #4 (convenience helper deferral)** — confirmed: the raw 2-call pattern (`capture(merchantAmount)` + `voidPayment()`) is the intended customer-facing surface for v0.3.0. A bundled `partialRefund(paymentInfo, refundAmount)` helper that wires the two calls is opt-in follow-up — not in this PR's scope.

CI watch starting now.

vraspar:
@A1igator — round 3 SDK Review (#issuecomment-4411258127) addressed in `4a70a18`. Apologies for missing this comment in the response cycle, only caught it on a re-pass.

All three findings landed in one commit:

1. **`presets.ts` L1207, L1271, L1356** — internal existence-tracking pipeline still used the bare `'releaseCondition'` string literal (deploy-plan entry name + `existsMap` lookup key + `exists.releaseCondition` consumer). The Round 2 commit (`29c0a33`) targeted `releaseConditionAddress` specifically via `replace_all`, so the suffix-less internal token escaped. All 3 sites now consistently use `'captureCondition'` / `exists.captureCondition`. Internal-only, no exported types affected.

2. **`presets.test.ts` L70, L101** — two `it(...)` titles in the marketplace describe block named `releaseCondition` even though the bodies asserted `capturePreActionCondition`. Updated to match. The parallel delivery-protection descriptions were already correct from `29c0a33`.

3. **Workspace `README.md` "Supported chains" L42-46** — listed 11 chains; `7ba81da` shrunk `x402rChains` to `{8453, 84532}` (canonical `commerce-payments@v1.0.0` coverage). README now lists `Base · Base Sepolia` with a one-line caveat about upstream coverage extension. Link to `packages/core/src/config` preserved.

Closing affirmations from the comment ("releaseConditionAddress on preset interfaces" + "stale partial-in-escrow refund migration in dispute-resolution.ts") match my view of state — both were already addressed in `29c0a33` / `eefae04` respectively.

CI watch starting now.

A1igator (approved):

PR #123, opened 2026-05-10
core+sdk+helpers+cli: authCapture wire format + autoCapture (PR 2/4 combined) [WIP]

vraspar:
## Summary

Combines PR 2 + PR 4 of the [authCapture SDK migration](https://github.com/BackTrackCo/x402r-notes/blob/main/plans/AUTHCAPTURE_SDK_MIGRATION.md). PR 4's `x402rDefaults()` builder + type re-exports are hard prerequisites for PR 2's autoCapture plumbing; splitting them would create rework. See the [combined-PR plan](https://github.com/BackTrackCo/x402r-notes/blob/main/plans/PHASE_2_AUTOCAPTURE_AND_WIRE_FORMAT.md) for the full task list.

Builds on PR 1 (#122).

## What ships

**Breaking**
- `@x402r/helpers` `forwardToArbiter` scheme filter `'commerce' → 'authCapture'`.
- `@x402r/evm` peerDep `>=0.0.3 → >=0.2.0-alpha.0 <0.3.0` on `@x402r/core`, `@x402r/helpers`. `@x402r/cli` dep bumped to the same range. The explicit range accepts both the alpha line and 0.2.0 final once published.
- `@x402r/cli` import switches from `@x402r/evm/commerce/client` (`registerCommerceEvmScheme`) to `@x402r/evm/authCapture/client` (`registerAuthCaptureEvmScheme`).
- `@x402r/core/payment` `toPaymentInfo()` now takes `PaymentInfoStruct` from `@x402r/evm` (was `EscrowPayload`, removed in `0.2.0-alpha.0`). Throws `ValidationError` on malformed wire input.

**New**
- `x402rDefaults(input) → AuthCaptureExtra` from `@x402r/helpers`. Quick-start builder with sensible defaults; only `captureAuthorizer` is required. `feeRecipient` defaults to `captureAuthorizer` (x402r deployment convention). Deadlines default to `now + 24h` / `now + 7d`. Fee policy `0`–`100` bps. Token domain `USDC` / `2`. Per-field JSDoc warns on production footguns.
- Wire-format type re-exports from `@x402r/helpers`: `AuthCaptureExtra`, `AuthCapturePayload`, `Eip3009Payload`, `Permit2Payload`, `PaymentInfoStruct` + payload type guards. Consumers building requirements don't need a direct `@x402r/evm` dep.
- `examples/scenarios/atomic-charge.ts` — single-tx `payment.charge()` flow with real ERC-20 balance-delta assertions.
- `examples/scenarios/partial-refund-flow.ts` — the new partial-refund pattern (`capture(partial)` then `voidPayment()`).

**Docs**
- `@x402r/core` README: opt-in note for direct-`@x402r/core` consumers on hook query scoping.
- `@x402r/sdk` README: trimmed "Refund & Dispute Flow" to 3 bullets (Evidence, Freeze roles, partial-capture recovery via `reclaim`). Deeper protocol semantics deferred to docs.x402r.org.
- `@x402r/helpers` README: quick-start snippet for `x402rDefaults`.

**Release tooling**
- Adds changeset; no manual version bumps. The changesets bot's "Version Packages" PR aggregates this with PR 1's pending bump at release time.

## Out of scope (PR 3)

Permit2 helper re-exports + `assetTransferMethod` dispatch in payment helpers + CLI flag.

## Test plan

- [x] Pre-commit hook green on every commit (`pnpm check && pnpm build && pnpm typecheck`)
- [x] `pnpm --filter @x402r/core test:unit` — 202 passed
- [x] `pnpm --filter @x402r/sdk test` — 196 passed
- [x] `pnpm --filter @x402r/helpers test` — 20 passed
- [x] `pnpm knip` — clean
- [x] All 4 scenarios run end-to-end on anvil: atomic-charge, happy-path-capture, partial-refund-flow, dispute-resolution. Real ERC-20 balance deltas asserted in the first three.
- [x] All 9 CI checks green (ABI Drift, Analyze, Build, Fork Tests, Lint & Audit, Publish Validation, Test, Typecheck, codecov/patch)

## Review rounds

- **Round 1 + 2 (A1igator):** 3 nits → addressed; Round 2 "Ready to ship."
- **Round 3 (vraspar, deep audit):** 9 issues addressed in `8543350..0b6ea76` — manual version bump revert, SDK `@x402r/evm` peerDep + re-export drop, core optional peerDep drop, atomic-charge balance-delta assertions, partial-refund example, sdk README trim, PR description rewrite, test-count correction.
- **Round 4 (vraspar):** 5 items. Items 1+2 (reclaim recovery note, happy-path-capture balance deltas) → `d4507d9`. Items 3+4 (workspace docs banned-word + marketing copy) → [BackTrackCo/docs#34](https://github.com/BackTrackCo/docs/issues/34). Item 5 (sdk README Action groups) → deferred per reviewer's own scope-creep call.
- **Round 5 + 6 (vraspar):** `x402rDefaults` redesign + JSDoc footgun warnings + `feeRecipient` defaults to `captureAuthorizer` → `0b73384`, `34fdec2`, `44690e5`.
- **Round 7 (vraspar):** 13 items, 4 commits `5303c10..16ebc31` — peerDep range widen, typed error on serialization boundary, helpers tests + comments + README snippet, MIGRATION.md deletion + sdk README restore + changeset trim. `toPaymentInfo` removal and `x402rDefaults` rename declined per user. Self-audit follow-up `55874b3` tightened the salt error matcher.


vraspar:
### Rubber-duck review of `6ae660f` (Task 2 commit)

**Verdict:** ship as-is.

**Issues ≥75:** none.

**Lower-scored observations (addressed in `6bf0818`):**

1. **(~65) Stale JSDoc reference** — `packages/core/src/payment/erc3009.ts:54` mentioned `@x402r/evm`'s `EscrowEvmScheme`, which is renamed to `AuthCaptureEvmScheme` in `0.2.0-alpha.0`. Fixed.
2. **(~60) Tautological test assertion** — `serialization.test.ts` "passes payer through unchanged" was meaningful when the old function extracted payer from `authorization.from`; with the new `PaymentInfoStruct` having `payer` baked in, the assertion was just `struct.payer === struct.payer`. Dropped (core unit tests now 200, was 201).
3. **(~55) Premature hard peerDep** — `packages/helpers/package.json` declares `@x402r/evm` as non-optional peer but doesn't yet import it. Non-issue: Task 3's `x402rDefaults()` will introduce the import in the next commit.

**Sweep:** zero remaining `EscrowPayload` / `commerce/client` / `commerce/server` / `registerCommerce` references in `packages/`, `examples/`, or `.changeset/`. No `examples/` file imports `@x402r/evm`, so nothing breaks against `0.2.0-alpha.0` from this commit.

**Forward-looking risks for Tasks 3-6:**
- `x402rDefaults()` must use wire-format field name `feeRecipient` (in `AuthCaptureExtra`), not on-chain `feeReceiver` (in `PaymentInfoStruct`). Already called out in the PHASE_2 plan.
- Atomic-charge example should use `isEip3009Payload` / `isPermit2Payload` type guards (exported from `@x402r/evm@0.2.0-alpha.0`) rather than untyped property access on the `AuthCapturePayload` discriminated union.

---
Generated with [Claude Code](https://claude.ai/code) using review-sdk skill (single-commit adaptation)

vraspar:
### Rubber-duck review of `c1f3069` (Task 3 commit)

**Verdict:** ship as-is.

**Issues ≥75:** none.

**Lower-scored observations:**

1. **(~65) viem dep redundant for `Address` type** — `packages/helpers/src/x402r-defaults.ts:2` imports \`Address\` from \`viem\`, adding a runtime dep to \`@x402r/helpers\`. Upstream \`AuthCaptureExtra\` types these fields as \`\\\`0x\${string}\\\`\` directly — no validation gained from the alias. **Will fix in next commit.**
2. **(~55) `PaymentInfoStruct` re-export of dubious value** — it's the on-chain Solidity struct, reconstructed by the facilitator from \`extra\` + salt. Merchants/payers don't construct it. The other four wire types (\`AuthCaptureExtra\`, \`AuthCapturePayload\`, \`Eip3009Payload\`, \`Permit2Payload\`) are the consumer-facing ones. Leaving in for now — \`toPaymentInfo()\` consumers might want it as input type — but flag for follow-up review.
3. **(~50) Duplicate-export collision risk** — both sdk and helpers re-export the same 4 type names + 4 type guards from \`@x402r/evm\`. Named imports resolve to same identity (fine). Wildcard imports (\`import * as\`) would collide. Worth a README sentence eventually, not blocking.

**Forward-looking risk for Task 4:** The atomic-charge example will likely **bypass `x402rDefaults()`**. The \`setup()\` helper in \`examples/shared/anvil-setup.ts\` constructs \`PaymentInfo\` (Solidity struct via \`@x402r/core\`) directly — not a \`PaymentRequirements\` with an \`extra\`. The merchant-as-own-facilitator path just calls \`payment.charge()\` directly. The example will likely build an \`extra\` with \`x402rDefaults({ autoCapture: true })\` purely as documentation of merchant intent, then call \`payment.charge()\` to do the actual atomic settlement. Code comments will need to make the demonstrative nature explicit.

**JSDoc accuracy:** verified — \`AuthCaptureServerScheme.enhancePaymentRequirements\` exists at \`node_modules/.../authCapture/server/index.d.mts:57\`.

**Sweep:** zero stale \`protocolFeeConfig\`-as-config-object references in the SDK. All 17 references correctly treat it as \`Address\` (matches \`packages/core/src/config/index.ts:39\` and contract ABI).

---
Generated with [Claude Code](https://claude.ai/code) using review-sdk skill (single-commit adaptation)

vraspar:
### SDK Review

Found 3 issues (axes: tests, dead code, example alignment):

1. **[Dead Code / Stale Docs After Refactor]** `EscrowServerScheme` reference in helpers README and JSDoc no longer resolves under `@x402r/evm@0.2.0-alpha.0`. The new peerDep exports `AuthCaptureServerScheme` (verified at `node_modules/.pnpm/@x402r+evm@0.2.0-alpha.0_*/dist/esm/authCapture/server/index.d.mts`); copy-paste from these blocks will fail to import.

   https://github.com/BackTrackCo/x402r-sdk/blob/638ac0af4f34d88192962278d59a3991f63dfff0/packages/helpers/README.md#L16-L18

   https://github.com/BackTrackCo/x402r-sdk/blob/638ac0af4f34d88192962278d59a3991f63dfff0/packages/helpers/src/forward-to-arbiter.ts#L20-L26

2. **[Tests]** `toPaymentInfo` test coverage dropped the `payer` assertion when migrating from `EscrowPayload` to `PaymentInfoStruct`, but the new struct has `payer` as a top-level field the function explicitly reads (`payer: struct.payer`). The remaining "passes through address and number fields unchanged" test only asserts `operator/receiver/token/feeReceiver/minFeeBps/maxFeeBps`. Dropping `payer:` from the return would silently pass. Add `expect(result.payer).toBe(baseStruct.payer)` to that test (or restore the dedicated case with the new fixture).

   https://github.com/BackTrackCo/x402r-sdk/blob/638ac0af4f34d88192962278d59a3991f63dfff0/packages/core/tests/serialization.test.ts#L27-L35

3. **[Example Alignment]** New `atomic-charge` scenario is wired into `examples/package.json` as `pnpm scenario:atomic-charge` and the file exists, but neither `examples/README.md` nor `examples/scenarios/README.md` lists it alongside `happy-path-capture` and `dispute-resolution`. Users browsing the docs won't discover the autoCapture demo.

   https://github.com/BackTrackCo/x402r-sdk/blob/638ac0af4f34d88192962278d59a3991f63dfff0/examples/scenarios/README.md#L11-L24

---

A1igator:
### SDK Review — Round 2

Found 0 new/unresolved issues since commit 638ac0a (axes: conventions, SDK design — not previously reviewed; tests, dead code, example alignment — re-checked for regressions in commit 16c2023).

- `EscrowServerScheme` references in `helpers/README.md` + `forward-to-arbiter.ts` JSDoc
- `payer` assertion in `serialization.test.ts` "passes through" case
- `atomic-charge` entry in `examples/scenarios/README.md`

**Still open:** none.
**New:** none.

Conventions clean (no ethers, no custom ABIs, no reinvented viem utils). SDK design is consistent with the operation-based architecture; wire-format type re-exports from both `@x402r/sdk` and `@x402r/helpers` are an intentional design choice with one-way deps. Upstream symbols (`AuthCaptureServerScheme`, `AuthCaptureExtra`, `registerAuthCaptureEvmScheme`, payload type guards) verified against `x402r-scheme/packages/evm/src/authCapture/`.

Ready to ship.

---

vraspar:
### SDK Review — Round 3 (deep audit)

Found 9 new/unresolved issues since commit `16c2023` (axes: release tooling, API consistency, SDK design, examples, docs, tests).

**Resolved since last round:** none (Round 2 found 0 issues).
**Still open from prior rounds:** none.
**New:**

1. **[Release tooling]** Manual version bumps in commit `638ac0a` fight the changesets workflow — will double-bump at release time.

   npm currently has `@x402r/core@0.2.0`, `@x402r/sdk@0.2.1`, `@x402r/helpers@0.2.1`, `@x402r/cli@0.2.0`. `main` is at 0.3.0/0.3.0/0.3.0/0.2.0 (PR 1's manual bump, never published). This PR bumps to 0.4.0/0.4.0/0.4.0/0.3.0 on top of that, and adds a changeset declaring `major × 4`.

   Standard changesets workflow: PRs add `.changeset/*.md` files only; a separate "Version Packages" PR runs `pnpm changeset version` to aggregate unreleased changesets and bump versions. When `changeset version` next runs against this branch, it sees TWO unreleased `major` changesets (PR 1's `sdk-authcapture-lift.md` + this PR's) and bumps again, pushing 0.4.0 → 0.5.0. The manual bumps become garbage that has to be reverted before publish.

   Fix: drop the version bumps in commit `638ac0a`. Keep only `.changeset/sdk-authcapture-autocapture-wireformat.md`. Versions in this PR should stay at main's values (0.3.0/0.3.0/0.3.0/0.2.0). Also worth a follow-up to clean up PR 1's manual bump.

   https://github.com/BackTrackCo/x402r-sdk/blob/16c2023200b76c3c43fb6f7f07866e11d5f2bab6/packages/core/package.json#L3

2. **[API consistency]** `@x402r/core` keeps `@x402r/evm` peerDep as `optional: true`, but `toPaymentInfo()` now takes `PaymentInfoStruct` as a public function parameter — type resolution breaks for direct-core consumers without the peerDep installed.

   `packages/core/package.json:55-60` retains `peerDependenciesMeta: { '@x402r/evm': { optional: true } }`. `packages/sdk/package.json` and `packages/helpers/package.json` dropped the same flag in commit `c1f3069` because they re-export evm types. Core also imports `PaymentInfoStruct` and uses it in the public signature at `packages/core/src/payment/serialization.ts:11`. Same type-resolution requirement, inconsistent declaration. Drop the `optional` flag on core.

   https://github.com/BackTrackCo/x402r-sdk/blob/16c2023200b76c3c43fb6f7f07866e11d5f2bab6/packages/core/package.json#L55-L60

3. **[Examples]** Missing runnable example for the partial-refund flow (`capture(partial) + voidPayment()`) — the central behavioral change of the migration has no scenario.

   The new partial-refund flow replaces the old single-tx `refundInEscrow(amount)` with two transactions: `capture(merchantAmount)` then `voidPayment()` returns the remainder. It's documented in inline comments at `examples/scenarios/dispute-resolution.ts:135-140` and `packages/sdk/README.md:124-126`, and tested at `packages/core/tests/partial-refund.fork.test.ts`, but has no runnable scenario. A merchant browsing examples to learn the new pattern won't find it.

   Add `examples/scenarios/partial-refund-flow.ts`: capture `merchantKeep` of `maxAmount`, void the remainder, assert payer received `(maxAmount - merchantKeep)` via balance delta. Migration plan flags this gap at `x402r-notes/plans/AUTHCAPTURE_SDK_MIGRATION.md:84`.

4. **[SDK design]** `@x402r/sdk` has zero internal usage of `@x402r/evm` — the peerDep was added purely for re-exports duplicated from `@x402r/helpers`.

   `rg "@x402r/evm" packages/sdk/src/` returns only `index.ts:47-60` (re-export lines). The SDK doesn't import a single value or type from `@x402r/evm` outside re-exports. The wagmi+viem precedent for re-exporting peer types rests on genuine internal usage (wagmi uses viem clients, transports, accounts internally). The SDK's case here is weaker — it's pure surface duplication.

   Two cleaner options:
   - Drop SDK re-exports. Keep wire-format types in `@x402r/helpers` only (helpers has real internal use via `x402rDefaults`). Consumers who want types import from helpers.
   - Drop both re-exports. Document `@x402r/evm` as the canonical import path in the changeset.

   If you keep the duplication, the changeset should explain why both packages surface the identical set.

   https://github.com/BackTrackCo/x402r-sdk/blob/16c2023200b76c3c43fb6f7f07866e11d5f2bab6/packages/sdk/src/index.ts#L47-L60

5. **[Examples]** `atomic-charge.ts` builds an `x402rDefaults({ autoCapture: true })` extra and then ignores it — example doesn't exercise the wire-format dispatch it advertises.

   `examples/scenarios/atomic-charge.ts:29-42` constructs `extra` then calls `payment.charge()` directly. The comment at line 16 calls the extra "decorative for documentation." The example demonstrates SDK-level `charge()` (which doesn't read `autoCapture`), not the wire-format flag's effect. Honest in comments, but a senior reviewer reading the example without the comments would conclude `autoCapture` is what triggered the atomic settlement, which is false.

   Fix options: (a) drop the decorative extra and replace with a one-line comment noting "in production, advertised in PaymentRequirements.extra," or (b) add a sibling example showing real facilitator dispatch that reads `extra.autoCapture` from a 402 response and chooses between `charge()` and `authorize()`.

   https://github.com/BackTrackCo/x402r-sdk/blob/16c2023200b76c3c43fb6f7f07866e11d5f2bab6/examples/scenarios/atomic-charge.ts#L29-L42

6. **[Examples]** `atomic-charge.ts` assertions are trivially true by construction — `capturableAmount === 0n` and `hasCollectedPayment === true` are contract-guaranteed post-`charge()`, catch nothing on the SDK side.

   `examples/scenarios/atomic-charge.ts:46-53`. Both invariants hold by definition after a successful `charge()` call. To be fair, `happy-path-capture.ts` has the same shallow pattern pre-existing — but perpetuating it instead of asserting balance deltas (payer balance ↓ by `PAYMENT_AMOUNT`, receiver balance ↑ by `(PAYMENT_AMOUNT - fee)`, fee receiver balance ↑ by `fee`) misses the chance to make the example a real regression guard.

   https://github.com/BackTrackCo/x402r-sdk/blob/16c2023200b76c3c43fb6f7f07866e11d5f2bab6/examples/scenarios/atomic-charge.ts#L46-L53

7. **[Docs]** `packages/sdk/README.md` "Refund & Dispute Flow" section (lines 110-126, 7 numbered bullets) is bloated relative to industry README norms — should be cut in this PR since the PR is already touching READMEs.

   Pre-existing on `main` but in scope for cleanup. Viem, wagmi, and Uniswap SDK READMEs keep package READMEs to quick-start + API surface; protocol edge cases live on docs sites or in `MIGRATION.md`. The current section explains in-escrow void auto-approval semantics, evidence multisig access control rules, post-escrow `ReceiverRefundCollector` staking, and partial-capture recovery — each ~3 sentences of protocol depth that belongs on docs.x402r.org.

   Suggested cuts (keep 2, move 5):
   - Keep: bullet 1 (Evidence 1:1 with RefundRequest — one-sentence contract fact), bullet 5 (Freeze roles — one-sentence role clarity).
   - Move to docs site or `MIGRATION.md`: bullets 2 (void auto-approve), 3 (evidence access control), 4 (post-escrow ReceiverRefundCollector), 6 (voidPayment OrCondition), 7 (partial-capture recovery / `reclaim`).

   Drops SDK README from 134 → ~110 lines.

   https://github.com/BackTrackCo/x402r-sdk/blob/16c2023200b76c3c43fb6f7f07866e11d5f2bab6/packages/sdk/README.md#L110-L126

8. **[Docs]** PR description is stale: still says "WIP DRAFT" and "Currently has 2 of ~6 commits. More to come." Actual state is 9 of 9 commits, all 6 logical units complete.

   Update the PR body before merge so reviewers don't misread state.

9. **[Tests]** PR description test plan claims 201 core unit tests pass; actual count is 200.

   Commit `6bf0818` dropped one tautological test in `serialization.test.ts` (correctly — the assertion became redundant when `payer` moved into `PaymentInfoStruct`). PR body's test plan was written before that commit and wasn't updated.

---

vraspar:
### Round 3 issues addressed

All 9 issues from [comment 4417015689](https://github.com/BackTrackCo/x402r-sdk/pull/123#issuecomment-4417015689) addressed in commits `8543350..0b6ea76`:

1. **Manual version bumps** → `8543350` reverts core/sdk/helpers to 0.3.0, cli to 0.2.0. Changesets bot's Version Packages PR will aggregate.
2. **Core optional peerDep flag** → `9be9a5d` drops `peerDependenciesMeta.@x402r/evm.optional` from `@x402r/core` (`toPaymentInfo()` uses `PaymentInfoStruct` in its public signature). `ba8f28a` follow-up drops the stale `knip.json` ignore that fell out.
3. **Missing `partial-refund-flow.ts`** → `b6a4bad` adds the scenario. Anvil run: payer recovers 400_000 remainder, receiver ↑ 597_000, fee ↑ 3_000 (50 bps of 600_000 merchant keep).
4. **SDK `@x402r/evm` peerDep + re-exports** → `c680fa3` drops both. SDK had zero internal evm usage — re-exports were pure surface duplication of `@x402r/helpers`. SDK net diff vs main is now zero in this PR; dropped from changeset frontmatter accordingly.
5. **atomic-charge decorative `x402rDefaults` extra** → `1b69fe8` drops the construction, replaces with a one-line comment about merchant intent in production `PaymentRequirements.extra`.
6. **atomic-charge tautological assertions** → same commit `1b69fe8` — replaces `capturableAmount === 0n` / `hasCollectedPayment === true` with real ERC-20 balance deltas. Anvil run: payer ↓ 1_000_000, receiver ↑ 995_000, fee ↑ 5_000.
7. **SDK README "Refund & Dispute Flow" bloat** → `0b6ea76` trims to 2 one-sentence bullets (Evidence 1:1, Freeze roles); moves the other 5 to new top-level `MIGRATION.md`.
8. **PR description stale** → rewritten via GraphQL mutation. No more "WIP DRAFT" framing.
9. **PR description wrong test count** → rewritten with accurate 200 / 196 / 14.

CI re-running on the head commit. Pre-commit hook green on every commit in this fix-forward set.

vraspar:
### SDK Review — Round 4

All 9 Round 3 issues resolved in commits `8543350..0b6ea76`. Found 4 new nits (all optional-before-merge or follow-up scope).

**Resolved since Round 3:**
- Issue 1 (manual version bumps) → `8543350`
- Issue 2 (core `@x402r/evm` `optional: true` flag) → `9be9a5d`, plus bonus knip corollary in `ba8f28a`
- Issue 3 (missing partial-refund example) → `b6a4bad`
- Issue 4 (SDK doesn't need `@x402r/evm`) → `c680fa3` — took the cleaner option (drop SDK re-exports entirely, keep helpers re-exports). Changeset frontmatter correctly updated to drop `@x402r/sdk` from `major` list since the SDK now has zero consumer-facing change.
- Issue 5 (atomic-charge decorative extra) → `1b69fe8` — replacement comment teaches that `autoCapture` is a facilitator-side dispatch decision, not an SDK operation
- Issue 6 (atomic-charge trivial assertions) → `1b69fe8` — real ERC-20 balance deltas (payer ↓ 1_000_000, receiver ↑ 995_000, fee ↑ 5_000)
- Issue 7 (SDK README bloat) → `0b6ea76` — surgical cut (kept bullets 1+5, moved 2/3/4/6/7 to `MIGRATION.md` at SDK repo root with proper section headers and README cross-link)
- Issue 8 (PR body "WIP DRAFT" rot) → PR body rewritten
- Issue 9 (test count claim 201 → 200) → PR body corrected

**Still open from prior rounds:** none.

**New (Round 4):**

1. **[Examples — in scope]** `examples/scenarios/partial-refund-flow.ts` has no recovery note for the half-completed-flow case.

   Grep returns zero matches for `reclaim`, `stuck`, `MIGRATION`, or `second tx` in the new example. The migration plan's backlog item 86 flags this as the merchant friction story: if `capture()` lands but `voidPayment()` fails (crash, gas exhaustion, key loss), the payer's remainder sits in escrow until `refundExpiry`, recoverable via `AuthCaptureEscrow.reclaim(paymentInfo)`. `MIGRATION.md` (added in `0b6ea76`) covers this — the example just needs to point at it.

   Fix: one-line comment at the start of the capture+void block: `// If the second tx fails, payer recovers the remainder via AuthCaptureEscrow.reclaim() after refundExpiry — see MIGRATION.md`.

   https://github.com/BackTrackCo/x402r-sdk/blob/0b6ea764e1a0bf9c7e1e7cba26669afbc24a7f78/examples/scenarios/partial-refund-flow.ts

2. **[Examples — in scope]** `examples/scenarios/happy-path-capture.ts` is now visibly shallow vs the new examples in the same folder.

   `atomic-charge.ts` and `partial-refund-flow.ts` (this PR) assert real ERC-20 balance deltas. `happy-path-capture.ts` only asserts `capturableAmount === 0n` and `hasCollectedPayment === true` — both contract-guaranteed post-capture, regression-guard value zero. The same `readBalance` helper from the new examples applies directly.

   Fix options: (a) upgrade `happy-path-capture` to assert payer/receiver/fee balance deltas — cheap, same pattern as the new files; (b) add a one-line comment documenting deliberate simplicity (e.g., `// Minimal happy-path; see atomic-charge.ts and partial-refund-flow.ts for balance-delta coverage`).

   https://github.com/BackTrackCo/x402r-sdk/blob/0b6ea764e1a0bf9c7e1e7cba26669afbc24a7f78/examples/scenarios/happy-path-capture.ts

3. **[Docs — workspace, follow-up]** `docs/index.mdx:7` uses the banned word "enables" — customer-facing on docs.x402r.org.

   Current: "It enables secure, reversible transactions on Base..."
   Rewrite: "x402r adds secure, reversible transactions with built-in buyer protection through smart contract escrow on Base."

   Out of this PR's scope (workspace `docs/` directory, not SDK), but worth filing as a separate docs cleanup ticket.

4. **[Docs — workspace, follow-up]** `docs/sdk/limitations.mdx:5` reads as marketing copy.

   Current: "The SDK provides full coverage of core payment flows..."
   Rewrite: "Core payment flows (authorization, capture, charge, refund, dispute resolution, evidence submission) are fully supported. This page documents known limitations."

   Same out-of-scope note as #3 — workspace `docs/` cleanup, not SDK PR responsibility.

**Lower priority (probably separate follow-up):**

5. **[Docs]** SDK README at 123 lines is improved but still verbose. "Action groups" section (`packages/sdk/README.md:39-49`) enumerates 7 groups across 40+ methods with no code examples. Industry pattern (viem, wagmi, uniswap-sdks) shows 1-2 code examples per group rather than enumeration. Scope creep for this PR — Round 3 fix already did the targeted trim the review asked for.

   https://github.com/BackTrackCo/x402r-sdk/blob/0b6ea764e1a0bf9c7e1e7cba26669afbc24a7f78/packages/sdk/README.md#L39-L49

---

vraspar:
### Round 4 issues addressed

- **Item 1** (partial-refund-flow missing reclaim recovery note) → `d4507d9` — one-line comment before the capture step pointing at MIGRATION.md for the half-completed-flow recovery story.
- **Item 2** (happy-path-capture shallow assertions) → same commit `d4507d9` — upgraded to real ERC-20 balance deltas mirroring atomic-charge.ts. Anvil run: payer ↓ 1_000_000, receiver ↑ 995_000, fee ↑ 5_000 (50 bps), matching atomic-charge numbers exactly.
- **Items 3+4** (workspace docs cleanup — banned word + marketing copy in `docs/`) → filed as [BackTrackCo/docs#34](https://github.com/BackTrackCo/docs/issues/34) per your call. Not bundled into #123.
- **Item 5** (SDK README "Action groups" still verbose) → deferred per your own scope-creep call.

All 4 scenarios still pass on anvil (happy-path-capture, atomic-charge, partial-refund-flow, dispute-resolution). Pre-commit hook green on `d4507d9`.

vraspar:
### SDK Review — Round 5

Two self-audit commits (`0b73384`, `34fdec2`) reviewed. The fee-delta tightening is a sharper take than I'd have suggested. The `x402rDefaults` defaults change is substantive and mostly correct, but 3 footguns deserve explicit JSDoc warnings.

**Resolved since Round 4:**
- Unposted micro-nit on `feeDelta >= 0n` → tightened to `> 0n` in `0b73384`. The commit message articulates a failure mode I missed (fee silently routes to receiver address; conservation passes; `>= 0n` passes; bug undetected). Strict-positive catches that class without coupling to `protocolFeeConfig` bps values. Better than my suggested exact-value approach.

**Still open from prior rounds:** none.

**New — JSDoc completeness on `x402rDefaults` defaults:**

The defaults change is a real improvement (8 required → 2 required), but 3 of the 6 defaults are high-risk for production copy-paste. The JSDoc says "For production, override the defaults that don't match your policy" without naming WHICH defaults are dangerous to leave alone.

1. **[Helpers — JSDoc]** `maxFeeBps: 100` default is a cap on what your facilitator can charge. If actual `protocolFeeConfig` exceeds 100 bps, every payment reverts on-chain.

   This is the single highest-risk default — the merchant has to know their facilitator's actual fee bps and won't discover the mismatch until production traffic. Worth a per-field JSDoc note: `"WARNING: caps the facilitator's allowed fee. If your facilitator's protocolFeeConfig exceeds 100 bps, payments revert. Override to match your facilitator's policy."`

   https://github.com/BackTrackCo/x402r-sdk/blob/34fdec2deca0fc644c224d9784b3d3246b3cb0d8/packages/helpers/src/x402r-defaults.ts#L20-L21

2. **[Helpers — JSDoc]** `captureDeadline: now + 1h` is short for asynchronous capture workflows. Merchants with worker queues, manual review steps, or batched capture will hit `void`-only state in production.

   Worth a JSDoc note: `"Defaults to 1 hour for quick-start; production async capture pipelines should override (24h+)."` Or escalate the default itself to `now + 24h` and document the contraction as a "tighten for atomic flows" note.

   https://github.com/BackTrackCo/x402r-sdk/blob/34fdec2deca0fc644c224d9784b3d3246b3cb0d8/packages/helpers/src/x402r-defaults.ts#L15-L16

3. **[Helpers — JSDoc]** `name: 'USDC'` + `version: '2'` are silent footguns for non-USDC tokens (EURC, PYUSD, DAI). EIP-712 typehash won't match the token's actual domain — signatures verify locally but fail on-chain submission.

   Worth a JSDoc note on `name`: `"Must match the token contract's EIP-712 domain name() exactly. Defaults to USDC; override for any other token (EURC, PYUSD, DAI, etc.) or signatures will fail verification on-chain."`

   https://github.com/BackTrackCo/x402r-sdk/blob/34fdec2deca0fc644c224d9784b3d3246b3cb0d8/packages/helpers/src/x402r-defaults.ts#L25-L28

4. **[Helpers — JSDoc]** Acknowledge the tension with `@x402r/evm`'s wire-format spec intent.

   The scheme's `AuthCaptureExtra` source comment explicitly says: "Fee-policy fields (minFeeBps, maxFeeBps, feeRecipient) are all required. No implicit defaults: a merchant who wants no minimum fee writes `minFeeBps: 0` explicitly. This forces fee policy to be a conscious choice on the wire and avoids 'did they mean 0 or did they forget?' ambiguity."

   This helper deliberately provides those implicit defaults at the SDK layer. That's a reasonable trade-off for a quick-start builder, but worth one line in the top JSDoc acknowledging it: `"Note: the underlying wire format requires fee fields explicitly to force conscious policy. This helper provides quick-start defaults at the SDK layer; production callers should override fee bps to match their facilitator's published policy."`

5. **[Helpers — JSDoc, lower priority]** Restore the audience note dropped from commit `c1f3069`.

   The original JSDoc named the audience: "merchants building requirements directly, or tests/examples that need a hand-built `extra`. The merchant's real-world flow merges facilitator's `/supported` extra into requirements via `AuthCaptureServerScheme.enhancePaymentRequirements`."

   The new JSDoc dropped that context. Production multi-facilitator merchants don't construct `extra` — they accept the facilitator's `/supported` advertisement. The helper's actual audience is merchant-as-facilitator (atomic-charge example pattern), single-tenant deployments, and tests. Worth one line restoring this so readers understand whether the helper applies to their setup.

---

vraspar:
### SDK Review — Round 6

Two follow-up items on `x402rDefaults`: one confirmation (no action), one suggestion that drops the required-field count from 2 → 1.

**Still open from prior rounds:** Round 5 items 1-5 (JSDoc completeness on defaults).

**New:**

1. **[Helpers — confirmation, no action]** The conditional spread on `autoCapture` and `assetTransferMethod` is correct as-is. Omitting when undefined is semantically identical to the wire-format spec's documented defaults (`autoCapture: false`, `assetTransferMethod: 'eip3009'`), avoids wire noise, and forces explicit opt-in for the risky atomic-charge path. Surfaced here so the audit record shows it was considered.

   https://github.com/BackTrackCo/x402r-sdk/blob/34fdec2deca0fc644c224d9784b3d3246b3cb0d8/packages/helpers/src/x402r-defaults.ts#L52-L57

2. **[Helpers — suggestion]** `feeRecipient` should default to `captureAuthorizer`. In the x402r deployment convention, the fee recipient is the same address as the capture authorizer (operator-and-fee-recipient are the same EOA in the facilitator setup). Encoding that convention drops the required-field count from 2 → 1:

   ```ts
   // After
   x402rDefaults({ captureAuthorizer: '0x...' })
   // feeRecipient defaults to captureAuthorizer
   ```

   This is a safer default than the `maxFeeBps: 100` / `name: 'USDC'` defaults flagged in Round 5: those have mismatch failure modes (revert, signature failure); `feeRecipient` defaulting to `captureAuthorizer` matches the deployment by construction. The override path (`feeRecipient: '0xTreasury...'`) covers the rare case of a separate fee treasury.

   The helper lives in `@x402r/helpers`, which is x402r-specific — encoding x402r deployment conventions is its job.

   Implementation:

   ```ts
   feeRecipient: input.feeRecipient ?? input.captureAuthorizer,
   ```

   With `feeRecipient` moved from required → optional in `X402rDefaultsInput`, JSDoc:

   ```ts
   /**
    * Address that receives the fee portion. Defaults to `captureAuthorizer`
    * (the x402r deployment convention). Override only if your facilitator
    * routes fees to a separate treasury.
    */
   feeRecipient?: \`0x\${string}\`
   ```

   Add one test: "feeRecipient defaults to captureAuthorizer when omitted." Existing override tests still hold because passing a value still wins.

   https://github.com/BackTrackCo/x402r-sdk/blob/34fdec2deca0fc644c224d9784b3d3246b3cb0d8/packages/helpers/src/x402r-defaults.ts#L11-L12

---

vraspar:
**Peer-dep range semantics — important**

The PR description says *"once `@x402r/evm@0.2.0` ships, the same constraint accepts the final."* That isn't how npm semver resolves caret-on-prerelease.

`^0.2.0-alpha.0` only matches further `0.2.0-alpha.N` prereleases on the same line. It will **not** automatically accept `0.2.0` (final). Once `@x402r/evm@0.2.0` publishes, consumers of `@x402r/core`, `@x402r/helpers`, and `@x402r/cli` will still be pinned to the latest `alpha` until we bump the range.

Two fixes:
- **a)** Widen now to `>=0.2.0-alpha.0 <0.3.0` in `packages/{core,helpers,cli}/package.json` so the final auto-applies.
- **b)** Keep `^0.2.0-alpha.0` and correct the PR description + changeset to flag the follow-up bump-at-release.

Either is fine, but the description as written is wrong.

vraspar:
**`serialization.ts` lacks input guards — important nit**

`packages/core/src/payment/serialization.ts:19,26` — `BigInt(struct.maxAmount)` and `BigInt(struct.salt)` throw `SyntaxError` on malformed wire input. Serialization is the natural enforcement boundary against bad upstream data.

Suggest:
- `salt: hexToBigInt(struct.salt)` (from viem) for the bytes32-hex.
- For `maxAmount`, keep `BigInt(...)` but wrap in a typed `try/catch` that throws a named error (`InvalidPaymentInfoStructError` or similar) — same pattern the CLI uses in `pay/policy.ts:46-49`.

Not a runtime bug today (assuming `@x402r/evm` always sends clean strings), but a real error-quality cliff if it doesn't.

vraspar:
**Address typing inconsistency**

`packages/helpers/src/x402r-defaults.ts:12,20` types `captureAuthorizer` and `feeRecipient` as `` `0x${string}` `` — structurally identical to viem's `Address` but missing the brand. The rest of the SDK (`packages/sdk/src/types.ts:36, 41-49, 62-69, ...`) uses `Address` consistently.

Either:
- Import `Address` from `viem` and use it — matches every other SDK input.
- Document explicitly that this mirrors `AuthCaptureExtra`'s upstream shape from `@x402r/evm@0.2.0-alpha.0` and that's intentional.

Same shape applies (separate scope) to `packages/cli/src/signers/{rpc.ts:29, key.ts:15}` which use the structural form after `isAddress()` narrows — the right cast is `as Address`.

vraspar:
**`x402rDefaults` has no downstream callers in this PR**

Exported and tested in isolation, but never demonstrated. `grep -rln "x402rDefaults" packages/ examples/` returns only `packages/helpers/{src,tests}`. An earlier commit had `atomic-charge.ts` calling it; Round 3 dropped that.

The "quick-start builder" should ship with at least one quick-start. Suggest adding a `x402rDefaults({ captureAuthorizer: '0x...' })` snippet to either:
- `packages/helpers/README.md` (right above `forwardToArbiter`),
- or one of the new scenarios (`atomic-charge.ts` is the natural fit).

vraspar:
**Builder name diverges from viem convention — non-blocking, raise before public surface locks**

`x402rDefaults()` reads like a constant ("the defaults") rather than a function that fills in defaults. Ecosystem convention for "typed config builder, no I/O, no state" is viem's `defineX` (`defineChain`, `defineTransport`).

Suggest renaming to `defineAuthCaptureExtra(input)` before the major version cuts — it names the output type and matches the `define*` precedent. Renaming after publish is another breaking change.

Alternates if you want to keep an `x402r` prefix: `x402rExtra({...})` (function-as-config, terse) or `createX402rExtra({...})` (matches `createWalletClient`, but slightly off since there's no I/O).

vraspar:
**`MIGRATION.md` is misnamed**

`MIGRATION.md` by convention (viem, wagmi v2, ethers v6) is **version-to-version upgrade steps** — old API → new API tables, removed exports, renamed args. This file is "non-obvious protocol behaviors," which is `ARCHITECTURE.md` or `docs.x402r.org` material. The preamble even acknowledges this: *"this file captures non-obvious refund-and-dispute behaviors and the authCapture wire-format migration paths"* — two documents fused.

Four of five sections (in-escrow void auto-approval, evidence access control, post-escrow `ReceiverRefundCollector`, voidPayment OrCondition) are protocol-behavior reference that overlaps `docs.x402r.org` and `x402r-notes/plans/AUTHCAPTURE_SDK_MIGRATION.md` — three-copy sync hazard.

Only **"Recovery if partial-capture's second tx never lands"** is load-bearing — it's a real footgun (`reclaim` not wrapped in the SDK) and isn't well-documented elsewhere.

Suggest: delete `MIGRATION.md`; move the `reclaim` paragraph into the `packages/sdk/README.md` dispute-flow section where someone actually hits the footgun. Update the link in `packages/sdk/README.md` accordingly.

vraspar:
**Changeset is heavy**

The changeset is 32 lines with nested bold-header sub-sections. Compare to viem's recent 2.x major changesets (typically 3-8 lines, one sentence per breaking item, no sub-headers). The `**New**` entry for `x402rDefaults` alone is 7 lines that copy JSDoc into changeset prose.

Suggest trimming to ~12 lines:
- Collapse each `**New**` bullet to one sentence. `x402rDefaults` becomes: `x402rDefaults(opts) → AuthCaptureExtra from @x402r/helpers — quick-start builder; only captureAuthorizer is required.`
- Drop the `**Out of scope (PR 3)**` section — internal PR sequencing isn't user-facing.
- Drop the `**Docs**` section — "trimmed protocol-depth, moved to MIGRATION.md" is a PR-review artifact.

vraspar:
**`isAuthCaptureExtra` as a validation oracle — nit**

`packages/helpers/tests/x402r-defaults.test.ts` uses `isAuthCaptureExtra(...)` in several "is valid" assertions (e.g. *"returns a value satisfying the AuthCaptureExtra type guard"*, *"produces a valid AuthCaptureExtra with only captureAuthorizer"*).

The guard does shape checks only (`typeof` on each field per `@x402r/evm@0.2.0-alpha.0`'s implementation) — it returns `true` for `feeRecipient: 'not-an-address'`, `minFeeBps: -5`, `captureDeadline: -1`. The exact-value assertions further down in the test file do verify field contents, so coverage holds, but the test names oversell what's being checked.

Either tighten the assertion (compare exact `extra` shape against an expected literal) or just rename to make the shape-vs-validity distinction explicit. Adding a one-line comment in the test explaining the shape-only nature of `isAuthCaptureExtra` would also work.

vraspar:
**Defaults handling is asymmetric — nit, design choice worth pinning**

`packages/helpers/src/x402r-defaults.ts:877-890` — most optional fields use `??` (so `feeRecipient: undefined` silently becomes `captureAuthorizer`), while `autoCapture` and `assetTransferMethod` use conditional-spread (so explicit-`undefined` is preserved as "omit from output").

This isn't a bug — it's the design (the two latter fields need the omit-from-wire signal to let the facilitator default win) — but it's a tripwire if someone later "tidies" the file to be uniform. Add a one-line comment above the conditional spread explaining why the asymmetry exists, so a future contributor doesn't simplify it away.

vraspar:
**SDK README trim dropped a clause that didn't migrate — small content loss**

`packages/sdk/README.md` — the "Refund & Dispute Flow" trim wasn't a clean move-to-`MIGRATION.md`. Bullets #1 (Evidence is 1:1 with RefundRequest) and #5 (Freeze roles) stayed inline. Bullet #1 was also *truncated* — the clause *"Evidence is required when refund is configured (`refundRequestEvidenceAddress` must be provided alongside `refundRequestAddress`)"* was dropped, and it doesn't appear in `MIGRATION.md` either. "Freeze roles" similarly isn't in `MIGRATION.md`.

The dropped clause is real integration guidance — a merchant configuring refund without the matching evidence address will fail at setup time. Either restore it in the SDK README under bullet #1, or move it to wherever the `RefundActions` interface is documented.

vraspar:
**Several `x402rDefaults` test cases are TS-redundant — nit, accompanies the `isAuthCaptureExtra` comment**

`packages/helpers/tests/x402r-defaults.test.ts` — the cases *"passes required fields through unchanged"* and *"overrides apply when fields are explicitly set"* are round-tripping inputs through `??` operators and asserting output === input. TypeScript already enforces this; the assertions verify nothing a refactor could break without `tsc` catching it first.

Cases worth keeping (these test real behavior): the four conditional-spread cases for `autoCapture` / `assetTransferMethod` (undefined → omit vs explicit-false → include), the four default-value cases (`feeRecipient` defaults to `captureAuthorizer`, deadline math, fee bps, USDC/2 token domain), and the `isAuthCaptureExtra` shape check.

Cases to either drop or merge into one tighter test: the two pure-passthrough ones. Suggest either deleting them or replacing with one snapshot-style test that compares `x402rDefaults(baseInput)` against an expected literal — that catches accidental field drops, missing fields, or extra fields in one assertion.

vraspar:
**AI-slop sentences in `MIGRATION.md` + changeset (specific quotes to delete)**

Independent of the broader structural calls on `MIGRATION.md` and the changeset (filed separately), these specific sentences don't carry weight and should be cut regardless of whether the files stay:

- **`MIGRATION.md`:** *"A typed wrapper is in the post-migration backlog."* — Roadmap chatter in user-facing docs. Either ship the wrapper or don't mention it. (If `MIGRATION.md` is deleted, this goes too. If kept, cut it.)
- **`MIGRATION.md` preamble:** *"Quick-start / API surface lives in the per-package READMEs; this file captures non-obvious refund-and-dispute behaviors and the authCapture wire-format migration paths."* — Self-acknowledged scope confusion. The file is announcing that it's two documents fused together.
- **Changeset:** *"x402r deployment convention — operator-and-fee-recipient are the same EOA"* — Parenthetical hand-wave inside a changeset bullet. Either it's a hard rule (then state it as such) or it's a convention (then it isn't changeset material).
- **Changeset:** *"Per-field JSDoc warns on the production footguns"* — Describing JSDoc *about* footguns is two levels removed. Either list the footguns or don't.
- **Changeset em-dashes:** 6 em-dashes in 32 lines, most acting as hedges. Replace with periods where you're using one to extend a complete thought ("regardless of any partial amount the payer requested", "or the arbiter approves").

vraspar:
**`toPaymentInfo` has no callers — intent check + suggestion to remove or relocate**

`grep -rln "toPaymentInfo"` across `x402r-sdk`, `arbiter-examples`, `docs`, `x402r-scheme`, and `tenjin` returns only the definition (`packages/core/src/payment/serialization.ts`), its tests, two barrel re-exports, and the changelog/README mentions. **Zero call sites** anywhere in the workspace.

**History.** Originally introduced in PR #31 as `helpers/src/bridge.ts` for e2e-test scaffolding ("http402 bridge utilities"). Moved into `core/utils` in commit `8b40d4c` (Feb 2026) with the rationale *"toPaymentInfo is a PaymentInfo converter, same category as `fromAbiPaymentInfo`, `parsePaymentInfo`, `toAbiPaymentInfo` already in core/utils. Helpers should only contain merchant-server utilities."* The original consumer (the e2e test) was wiped in `d0f4da0` ("clean slate"), and the sibling converters (`fromAbiPaymentInfo`, `parsePaymentInfo`, `toAbiPaymentInfo`) that justified the move into `core/utils` also no longer exist in core. So `toPaymentInfo` is now an isolated leftover whose entire context — caller, sibling utils, original rationale — has been deleted around it.

**What it does, in simple terms.** Thin deserializer for one boundary. Wire form (`PaymentInfoStruct` from `@x402r/evm`, JSON-safe strings for uint120/uint256) → runtime form (`PaymentInfo` from `@x402r/core`, real `bigint`s) that viem-flavored actions accept. The boundary is real (HTTP `/verify` callback handlers, future `payment.charge(paymentPayload)` overloads, typed arbiter receivers), but no current code in the workspace crosses it via this function — the arbiter example consumes `PaymentPayload` from upstream `@x402/core/types` and doesn't reach into x402r runtime form; the SDK scenarios construct `PaymentInfo` directly with `bigint` literals (`examples/shared/anvil-setup.ts:224-237`).

**Peer-dep implication.** `grep "@x402r/evm" packages/core/src/` returns only `serialization.ts` (real `import type { PaymentInfoStruct }`) and `erc3009.ts` (JSDoc-only reference, no actual import). If `toPaymentInfo` is removed, the peerDep has no remaining justification in `@x402r/core` — drop it from `packages/core/package.json` along with the matching devDep and the `@x402r/evm` snapshot in `pnpm-lock.yaml`. The whole "optional → required" peerDep tightening (Round 3 issue #2) becomes moot — we just don't have the peer dep in core at all.

**Three options:**
1. **Document and keep** — add a `@x402r/core` README snippet showing the boundary use case. Keep the peerDep. Lowest churn, but doesn't address "exported with no consumer."
2. **Move back to `@x402r/helpers`** (where it started in PR #31). Drops core's `@x402r/evm` peerDep entirely. Fits helpers' merchant-server charter. Reverses commit `8b40d4c`'s categorization, but that rationale is weak now — the sibling converters it grouped against don't exist anymore.
3. **Remove entirely.** Drop the function, drop core's `@x402r/evm` peerDep. Cleanest. If a real external consumer surfaces, add it back with docs. Most aggressive.

Recommendation: option 2 or 3. Both eliminate the orphan-export footprint and simplify `@x402r/core`'s dep graph. Option 2 keeps the utility available to integrators who land on the boundary; option 3 is cleanest if you'd rather not have the function at all.

vraspar:
### Round 7 issues addressed

13 items, 4 commits (`5303c10..16ebc31`) + 2 architectural decisions:

**`5303c10` — deps: widen @x402r/evm range**
- **C1** Peer-dep semantics → widened to `>=0.2.0-alpha.0 <0.3.0` on core/helpers (peerDeps) and cli (dep). devDeps stay pinned at exact `0.2.0-alpha.0`. Now unambiguously accepts both alpha line + 0.2.0 final once published.

**`9175356` — core: typed errors on serialization.ts**
- **C2** Input guards → `BigInt(struct.maxAmount)` wrapped in try/catch throwing `ValidationError` with the offending string. `BigInt(struct.salt)` swapped for viem's `hexToBigInt` (typed errors on malformed bytes32). Tests 200 → 202 (added: non-numeric maxAmount throws, malformed salt hex throws).

**`57cd20c` — helpers: drop TS-redundant tests, document type + defaults asymmetry**
- **C3** Address typing → kept `` `0x${string}` `` (mirrors `AuthCaptureExtra` in `@x402r/evm@0.2.0-alpha.0`, avoids reintroducing viem dep on helpers). One-line JSDoc note explains the choice.
- **C4** x402rDefaults README snippet → added "Building PaymentRequirements.extra" section to `packages/helpers/README.md` with the minimal `x402rDefaults({ captureAuthorizer })` call shape.
- **C8** `isAuthCaptureExtra` framing → one-line comment in tests explaining the guard is shape-only (typeof per field).
- **C9** Defaults asymmetry → one-line comment above the conditional-spread block explaining why `autoCapture`/`assetTransferMethod` use omit-when-undefined vs `??` for fields with fallback defaults.
- **C11** Drop TS-redundant tests → removed the two pure-passthrough cases. Helpers test count 22 → 20.

**`16ebc31` — docs: delete MIGRATION.md, restore sdk README clauses, trim changeset**
- **C6** MIGRATION.md fate → deleted entirely. By viem/wagmi/ethers convention `MIGRATION.md` is version-upgrade steps, not protocol reference. The reclaim recovery paragraph (only load-bearing content) moved back to `packages/sdk/README.md` bullet #3.
- **C10** SDK README truncated clause → restored "Evidence is required when refund is configured..." in bullet #1.
- **C7** Changeset trim → reduced 33 → 20 lines (target was ~12-14; kept blank-line separators between sections for viem-style readability). Dropped `**Docs**` and `**Out of scope**` sections.
- **C12** AI-slop quotes → removed: "A typed wrapper is in the post-migration backlog" (from reclaim text), "x402r deployment convention — operator-and-fee-recipient are the same EOA" (from changeset), "Per-field JSDoc warns on the production footguns" (from changeset). Em-dashes-as-hedges replaced with periods.

**Decisions (no commit):**
- **C5** Rename `x402rDefaults()` → declined. Name accurate now defaults exist; rename would require coordinated update across migration plan + changeset + PR description. The `x402r`-prefix is intentional branding.
- **C13** `toPaymentInfo` removal → deferred. Function stays in `@x402r/core` along with the `@x402r/evm` peerDep. Tracked for a separate follow-up.

**Verification:**
- `@x402r/core` tests: 202 passed (was 200).
- `@x402r/sdk` tests: 196 passed (unchanged).
- `@x402r/helpers` tests: 20 passed (was 22).
- All 4 scenarios pass on anvil (atomic-charge, happy-path-capture, partial-refund-flow, dispute-resolution).
- biome / typecheck / build / knip all clean.
- CI green on head (Fork Tests + ABI Drift still running last check; full pass expected).

A1igator:
### SDK Review — Round 8

Found 1 new issue since commit `34fdec2` (axes: error handling).

**Resolved since last round (Round 5/6 `### SDK Review` + author-labeled inline Round 7):**
- Round 5 items 1-5 (JSDoc completeness on `x402rDefaults` defaults: `maxFeeBps` revert warning, `captureDeadline` async-vs-atomic framing, `name`/`version` token-domain mismatch warning, restored audience note, wire-spec tension acknowledged) resolved in `44690e5`. Also escalated `captureDeadline` default 1h → 24h.
- Round 6 item 2 (`feeRecipient` defaults to `captureAuthorizer`, dropping required-field count 2 → 1) resolved in `44690e5`.
- Round 7 inline C1 (peer-dep semantics; widened to `>=0.2.0-alpha.0 <0.3.0`) resolved in `5303c10`.
- Round 7 inline C2 (input guards on `serialization.ts`) partially resolved in `9175356` (see new issue below).
- Round 7 inline C3 (address typing on `` `0x${string}` ``) resolved in `57cd20c` with explanatory JSDoc.
- Round 7 inline C4 (`x402rDefaults` README snippet) resolved in `57cd20c`.
- Round 7 inline C6 (`MIGRATION.md` misnamed; reclaim paragraph moved back to `packages/sdk/README.md`) resolved in `16ebc31`.
- Round 7 inline C7 (changeset trim, 33 → 20 lines) resolved in `16ebc31`.
- Round 7 inline C8 (`isAuthCaptureExtra` shape-only framing) resolved in `57cd20c`.
- Round 7 inline C9 (defaults-handling asymmetry, one-line comment) resolved in `57cd20c`.
- Round 7 inline C10 (truncated "Evidence is required..." clause) restored in `16ebc31`.
- Round 7 inline C11 (TS-redundant tests dropped, 22 → 20) resolved in `57cd20c`.
- Round 7 inline C12 (AI-slop quotes from `MIGRATION.md` and changeset) resolved in `16ebc31`.

**Author-acknowledged, no action (correctly):**
- Round 7 inline C5 (rename `x402rDefaults` to viem-style `defineX`) declined as intentional branding.
- Round 7 inline C13 (`toPaymentInfo` orphan) deferred to a separate follow-up.

**New:**

1. **[Error handling]** `salt` path in `toPaymentInfo` skipped the typed-error wrap that `maxAmount` got. The wire-format boundary now produces asymmetric errors at the same call site: malformed `maxAmount` throws `ValidationError("toPaymentInfo: invalid maxAmount '...'")` with the offending string and a `cause` chain, while malformed `salt` produces a bare native `SyntaxError("Cannot convert 0xZZZ to a BigInt")` from V8 with no `toPaymentInfo` context.

   Commit `9175356` ("typed errors on serialization.ts wire-format boundary") and commit `55874b3` ("symmetric in behavior") both frame the behavior as symmetric. It is not. Viem's `hexToBigInt` is `BigInt(hex)` with optional size checking (verified at `node_modules/.pnpm/viem@2.47.1_*/node_modules/viem/_esm/utils/encoding/fromHex.js`), so it passes V8's native BigInt SyntaxError straight through. A stack trace originating in viem internals reads as a library bug, not a contract-data validation failure at the SDK boundary. Test matcher `/Cannot convert/i` is also engine-string-coupled rather than module-name-coupled.

   Mirror the maxAmount block:

   ```ts
   let salt: bigint
   try {
     salt = hexToBigInt(struct.salt)
   } catch (err) {
     if (err instanceof SyntaxError) {
       throw new ValidationError(
         `toPaymentInfo: invalid salt '${struct.salt}' (expected 0x-prefixed bytes32 hex)`,
         { cause: err },
       )
     }
     throw err
   }
   ```

   Tighten the salt test matcher to `/salt/` for symmetry with the existing `/maxAmount/` matcher.

   https://github.com/BackTrackCo/x402r-sdk/blob/55874b342c72a2df8f41250a01664413d191ed69/packages/core/src/payment/serialization.ts#L41

---

vraspar:
### Round 8 issue addressed

A1igator's finding was sharp — my "symmetric in behavior" framing on `9175356`/`55874b3` was wrong. The `maxAmount` path threw `ValidationError("toPaymentInfo: invalid maxAmount '...'")` with cause chain; the `salt` path passed V8's bare `SyntaxError` through viem's `hexToBigInt` (no wrapping at the viem layer — verified `BigInt(hex)` + size check at `node_modules/.../viem/_esm/utils/encoding/fromHex.js`).

Fixed in `dffe896`:

- `serialization.ts`: wrap `hexToBigInt(struct.salt)` in the same try/catch pattern as `maxAmount` — `ValidationError("toPaymentInfo: invalid salt '...'")` with cause chain.
- `serialization.test.ts`: tighten salt matcher from engine-coupled `/Cannot convert/i` to `/salt/` for symmetry with the existing `/maxAmount/` matcher.

Tests still at 202 passing. Both wire-format-boundary error paths now produce SDK-recognizable failures rather than library-internal stack traces.

A1igator (approved):

PR #124, opened 2026-05-13
core+helpers: relocate toPaymentInfo (PR 4 of authCapture migration)

vraspar:
## Summary

Move `toPaymentInfo` from `@x402r/core/payment/serialization` to `@x402r/helpers`. Drop `@x402r/evm` peerDep+devDep from `@x402r/core` — nothing else in core consumed it.

`@x402r/core` becomes the cleanest layer in the workspace: zero `@x402*` deps, just `viem`.

PR 4 of 4 in the authCapture SDK migration (see `x402r-notes/plans/AUTHCAPTURE_SDK_MIGRATION.md`). PR 2/4 combined merged in #123; this is the deferred relocation surfaced in Round 7.

## Breaking

- `@x402r/core` no longer exports `toPaymentInfo` / `ToPaymentInfoReturnType`. Import from `@x402r/helpers` instead.
- `@x402r/core` no longer declares `@x402r/evm` as a peerDep.

## Side-effect changes (not in the relocation plan, surfaced during build)

- `@x402r/helpers` now declares `viem` as a direct dep (the relocated function imports `hexToBigInt`). Matches the workspace convention used by `@x402r/core` and `@x402r/sdk`.
- The moved test inlines `TEST_ADDRESSES` (5 hex strings) since `@x402r/helpers/tests` has no shared fixtures file; creating one would be YAGNI.

## Test plan

- [x] Unit tests pass: 197 core / 196 sdk / 25 helpers (total 418 — same as baseline, the 5 serialization tests just moved packages).
- [x] `pnpm knip` clean.
- [x] `pnpm check && pnpm build && pnpm typecheck` green.
- [x] `rg "@x402r/evm" packages/core/` → only the JSDoc-only mention in `erc3009.ts:55`.
- [x] `rg "toPaymentInfo" packages/ examples/` → only the helpers source + test + barrel.
- [x] `git log --follow packages/helpers/src/serialization.ts` shows pre-rename history.
- [ ] CI checks all green: ABI Drift, Analyze, Build, Fork Tests, Lint & Audit, Publish Validation, Test, Typecheck, codecov/patch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #126, opened 2026-05-14
core+helpers: PaymentInfo namespace + reconstructPaymentInfoWire

vraspar:
## Summary

Adds the `PaymentInfo` namespace to `@x402r/core` (re-exported from `@x402r/sdk`) and adds `reconstructPaymentInfoWire` to `@x402r/helpers`. Together they bridge the authCapture wire format end-to-end: the helper produces the JSON-form wire record at settle time, `PaymentInfo.fromWire` hydrates it back to bigints right before SDK actions.

This is a reshape from the earlier helpers-only version of this PR. See `x402r-notes/plans/SDK_PAYMENTINFO_NAMESPACE.md` and `SDK_PAYMENTINFO_ERGONOMICS_DECISION.md` for the design rationale.

## API at a glance

```ts
import { PaymentInfo } from '@x402r/sdk'

// arbiter / worker / refund endpoint
const info = PaymentInfo.fromWire(req.body.paymentInfoWire)
await client.payment.capture(info, info.maxAmount)
```

```ts
// merchant hook at settle time
import { reconstructPaymentInfoWire } from '@x402r/helpers'
import { PaymentInfo } from '@x402r/sdk'

resourceServer.onAfterSettle(async (context) => {
  const wire = reconstructPaymentInfoWire(context)
  await db.jobs.create({ paymentInfoWire: wire, ... })  // JSON-safe persist
  // OR in-process:
  const info = PaymentInfo.fromWire(wire)
  await client.payment.capture(info, info.maxAmount)
})
```

## What's new

- **`PaymentInfo` namespace** in `@x402r/core` (re-exported from `@x402r/sdk`) — same identifier serves as the bigint type *and* the converter namespace via the standard TypeScript type+const colocation pattern (same shape as `Date`, `Buffer`).
- **`PaymentInfoWire` type** in `@x402r/core` — JSON-safe shape, derived from `PaymentInfo` via a new `AbiPrimitiveToWire<T>` helper. Stays in sync with the bigint shape at compile time when the ABI changes.
- **`reconstructPaymentInfoWire(context)`** in `@x402r/helpers` — builds the wire form from a verified `SettleResultContext`. Handles the 6 wire→struct field renames and the EIP-3009 vs Permit2 branch internally.

## Breaking

- **`@x402r/helpers`: `toPaymentInfo` removed.** Use `PaymentInfo.fromWire` from `@x402r/sdk` instead. **Arbiters and standalone workers can now drop `@x402r/helpers` entirely.**
- **`@x402r/helpers`: `reconstructPaymentInfoStruct` renamed to `reconstructPaymentInfoWire`.** Return type renamed from `PaymentInfoStruct` to `PaymentInfoWire`.
- **`@x402r/helpers`: `forwardToArbiter` POST body field renamed** from `paymentInfoStruct` to `paymentInfoWire`. Arbiters should switch `req.body.paymentInfoStruct` → `req.body.paymentInfoWire` and use `PaymentInfo.fromWire(...)`.

## Dependency story (the load-bearing design point)

`PaymentInfo` and the converters live in `@x402r/core` without re-introducing the `@x402r/evm` peerDep that PR #124 worked to remove. The trick: `PaymentInfoWire` is **derived from the contract ABI** that `@x402r/core` already owns, not imported from `@x402r/evm`. Core stays scheme-agnostic; `@x402r/evm` ships unchanged (important — it's being upstreamed to x402 spec).

`@x402r/evm`'s `PaymentInfoStruct` keeps existing as an unchanged hand-written interface for any direct consumers. The two types are structurally identical but defined independently — no cross-package coupling.

## Caveats

- **`type + const` sharing a name** is a standard TS pattern (`Date`, `Buffer`, `Promise`, `Array` all do it) but requires the barrel re-export to use `export { PaymentInfo }` without the `type` keyword to forward both meanings in one declaration.
- **`@x402r/evm`'s `PaymentInfoStruct` is hand-written** — if its shape ever drifts from core's ABI-derived `PaymentInfoWire`, the two would no longer be interchangeable. The fix when that happens is upstream's responsibility (or a small change in evm to align).
- **`amount` arg to `payment.capture`/`refund` stays bigint.** It's a separate semantic decision (capture amount vs maxAmount), not part of `PaymentInfo`. Callers typically pass `info.maxAmount` directly (already a bigint after `fromWire`) — no extra `BigInt(...)` call needed in the common case.

## Test plan

- [x] 216 core / 196 sdk / 28 helpers tests pass (helpers shrinks by 5 from removing `toPaymentInfo`; core gains 9 new namespace tests including round-trip)
- [x] `pnpm knip` clean
- [x] `pnpm check && pnpm build && pnpm typecheck` green
- [x] `PaymentInfo` works as both a type annotation and a value namespace (`info: PaymentInfo` + `PaymentInfo.fromWire(...)`)
- [x] `forwardToArbiter` body field verified as `paymentInfoWire` in tests
- [ ] CI checks all green: ABI Drift, Analyze, Build, Fork Tests, Lint & Audit, Publish Validation, Test, Typecheck, codecov/patch

## Out of scope

- Updating `arbiter-examples` to consume the new shape — tracked separately as the "arbiter-examples staleness" backlog item.
- Generic ABI codec for action-side conversion (the friend's original "approach B" proposal). Decision doc parks it under N≥5 wire-crossing structs trigger; today we have 1.


A1igator:
### SDK Review

No issues found at the flag threshold (axes: tests, conventions, dead code, example alignment, SDK design).

**Pass notes:**
- Both EIP-3009 and Permit2 `preApprovalExpiry` branches covered; all three `ValidationError` paths exercised; `globalThis.fetch` saved/restored in every test including the new skips-POST regression guard.
- `reconstructPaymentInfo` fits the companion pattern with `toPaymentInfo`; two-step API justified by the worker-boundary persistence example in README.
- `forwardToArbiter` already gates on `result.success` and `scheme === 'authCapture'` before reconstruction, so the new throw paths only fire on genuinely malformed contexts.
- Changeset `major` bump is correct for the `paymentPayload` to `paymentInfo` body rename; migration note covers the consumer-side switch.

**Informational only (below threshold, not blocking):**
1. `reconstructPaymentInfo` JSDoc lists "wrong scheme" as a throw case, but the function checks `extra` shape via `isAuthCaptureExtra` rather than reading `requirements.scheme`. End-to-end behavior is correct because the structural guard is the effective discriminator, but tightening the doc (or adding an explicit scheme check) would remove the discrepancy.
2. `salt` is returned without hex-prefix validation; the error surfaces one hop later inside `toPaymentInfo` via `BigInt(...)`. Same pattern as neighboring fields, fine as-is.

---

vraspar:
Addressed informational note #1 in 22f3da1 — tightened the JSDoc so the listed throw conditions match the structural checks the implementation actually performs (`isAuthCaptureExtra` / `isAuthCapturePayload` / payer presence), with the non-authCapture-scheme case called out as a parenthetical under the first bullet since that's the common real-world trigger.

Note #2 (`salt` hex validation deferred to `toPaymentInfo`'s `BigInt(...)` call) left as-is per the review — same pattern as neighboring fields.

A1igator (dismissed):

vraspar:
**PR reshaped (force-push, 6 commits → 4)** to ship the namespace API (option C from the design discussion) instead of the original helpers-only `toPaymentInfo` form. Net result: arbiters and standalone workers can now drop `@x402r/helpers` entirely and just `import { PaymentInfo } from '@x402r/sdk'`.

**Why force-push vs follow-up PR:** the rework touches the same surface as the original (helpers' `toPaymentInfo` removal, `forwardToArbiter` body field rename) so a follow-up PR would have to undo + redo the same lines. One amended PR is cleaner. Pre-merge, no published consumer impact.

**Dep story** (the load-bearing design point, expanded in PR body):
- `PaymentInfo` and `PaymentInfoWire` live in `@x402r/core`, both derived from the contract ABI core already owns
- **No `@x402r/evm` peerDep added to core** — PR #124's cleanup is preserved
- `@x402r/evm` ships unchanged (important — being upstreamed to x402 spec)

**Caveats worth noting:**
- The TS `type + const` colocation for `PaymentInfo` requires the barrel to use `export { PaymentInfo }` (no `type` keyword) so both meanings re-export in one declaration. Standard pattern (`Date`, `Buffer`, etc.) but worth knowing about.
- `@x402r/evm`'s hand-written `PaymentInfoStruct` stays for back-compat; structurally identical to core's ABI-derived `PaymentInfoWire`. If they ever drift (e.g., contract ABI change without an evm bump), the alignment breaks — but that's an upstream responsibility.
- `amount` arg to actions stays `bigint`. Common case `payment.capture(info, info.maxAmount)` requires no extra conversion — `info.maxAmount` is already a bigint after `fromWire`.

Long-form rationale in `x402r-notes/plans/SDK_PAYMENTINFO_NAMESPACE.md` and `SDK_PAYMENTINFO_ERGONOMICS_DECISION.md` (the latter's addendum covers the decision tree under varying struct counts and the trigger conditions for the deferred codec option).

Local verification: 216/196/28 tests, knip clean, check/build/typecheck green. CI re-running on the force-push.

A1igator:
### SDK Review — Round 2

No new issues at the flag threshold since 520840493 (axes: tests, conventions, dead code, example alignment, SDK design).

**Resolved since round 1:**
- `reconstructPaymentInfo` JSDoc that listed "wrong scheme" as a throw case — JSDoc on the renamed `reconstructPaymentInfoWire` now accurately describes the `isAuthCaptureExtra` / `isAuthCapturePayload` / missing-payer guards.

**Still open from round 1:**
- `salt` hex-prefix is not pre-validated; the gap moved with the function (was in `toPaymentInfo`, now in `PaymentInfo.fromWire`). Symmetric `maxAmount` decimal-only check is also absent. `BigInt('1234')` accepts decimal salt as 1234n and `BigInt('0x10')` accepts hex maxAmount as 16n — both pass despite the error messages claiming the opposite format. Same severity as before, not blocking.

**Pass notes:**
- `PaymentInfo` type-and-const sharing a name follows the documented Date/Buffer idiom; single `export { PaymentInfo }` re-exports both meanings.
- `AbiPrimitiveToWire<T>` derives `PaymentInfoWire` from `PaymentInfo` at the type level, so contract ABI changes propagate to both shapes — no parallel hand-maintained wire type.
- Architectural placement justified by the changeset's "Why these live where they do" section: conversion is scheme-agnostic (core), reconstruction is scheme-specific (helpers). Arbiters and standalone workers can drop the `@x402r/helpers` dep entirely now.
- Body field rename `paymentInfoStruct → paymentInfoWire` is correctly flagged as breaking in the changeset with an explicit migration path; the regression-guard test (`skips POST and calls onError when reconstructPaymentInfoWire throws`) covers the failure-to-reconstruct path.
- 9 tests on `PaymentInfo.fromWire`/`.toWire` covering both directions, round-trip, hex-edge for small salts, and both `ValidationError` branches; all `globalThis.fetch` references in `forward-to-arbiter.test.ts` are saved and restored in `try`/`finally`.

**Informational only (below threshold, not blocking):**
1. `packages/core/src/types/index.ts:49` — JSDoc on `fromWire` says "Convert a JSON-form `PaymentInfoStruct` (or `PaymentInfoWire`)"; the signature only accepts `PaymentInfoWire`. Stale `PaymentInfoStruct` reference from the rename. Tightening to just `PaymentInfoWire` removes the ambiguity (the line-25 comment already explains the historical equivalence with `@x402r/evm`'s `PaymentInfoStruct`).
2. `packages/helpers/tests/forward-to-arbiter.test.ts:96` — comment reads "the reconstructed PaymentInfoStruct under `paymentInfoWire`"; should say `PaymentInfoWire` for consistency with the asserted field name.

---

A1igator:
### Naming nit: `PaymentInfoJSON` / `toJSON` / `fromJSON`

Round-2 review didn't surface this, but worth raising on its own. The `Wire` naming is the one thing left in the new API that reads as x402-protocol-specific when the shape itself isn't.

Two reasons to consider renaming before the major lands:

1. **Built-in symmetry.** The changeset already pitches the namespace as "same shape as `Date`, `Buffer`." Both expose `toJSON()` for their JSON-safe form. Naming consistency with the built-ins matches the rationale already in the changeset.

2. **Precision.** `toJSON`/`fromJSON` describes exactly what the transform does (bigint↔string for JSON-safety). `Wire` is broader and ambiguous; it could read as HTTP wire, Protobuf wire, or x402 wire. The new name removes that ambiguity.

No conflict with `JSON.stringify`'s `toJSON` auto-call: the method is static on the namespace const, not an instance method on `PaymentInfo` values, so the auto-invocation path doesn't trigger.

Concrete rename:
- `type PaymentInfoWire` → `type PaymentInfoJSON`
- `PaymentInfo.toWire` / `PaymentInfo.fromWire` → `PaymentInfo.toJSON` / `PaymentInfo.fromJSON`
- `AbiPrimitiveToWire<T>` → `AbiPrimitiveToJSON<T>`
- `reconstructPaymentInfoWire` → `reconstructPaymentInfoJSON`
- POST body field `paymentInfoWire` → `paymentInfoJson`

Mechanical search/replace, no runtime change. Easier now than after the major ships.

---

vraspar:
Addressed in 4bafa7b — tightened the salt validation error message to say 'uint256 (decimal or 0x-prefixed hex)' since the ABI declares salt as uint256, not bytes32 (BigInt() accepts both forms). Also updated PR title to reflect that the load-bearing change is the core namespace, not just the helpers piece.

Non-blocking note #2 (PaymentInfoStruct's salt is `0x${string}` while PaymentInfoWire's is plain `string`): true at the type level, structurally compatible at runtime since both flow through BigInt(). Acceptable as-is; would need an x402r/evm change to align which is out of scope for this PR.

Non-blocking note #4 (AbiPrimitiveToWire's array recursion untested): kept as defensive future-proofing for future structs with array fields. Will get exercise when/if a contract struct uses arrays.

vraspar:
Addressed round 2 informational items in ea47a05:

- **Stale `PaymentInfoStruct` in JSDoc** (`packages/core/src/types/index.ts:49`) — dropped the `'PaymentInfoStruct (or PaymentInfoWire)'` phrasing; signature only accepts `PaymentInfoWire`.
- **Stale `PaymentInfoStruct` in test comment** (`packages/helpers/tests/forward-to-arbiter.test.ts:96`) — now reads `PaymentInfoWire` to match the asserted field name.
- **`maxAmount` error message vs BigInt leniency** — loosened to `'expected uint256 (decimal or 0x-prefixed hex)'` to match `BigInt()`'s actual behavior and align with salt's already-corrected wording. Chose to match the runtime rather than tighten validation; both wire formats are valid uint256 representations and rejecting hex would be arbitrary.

**On the `Wire` → `JSON` naming proposal** (separate comment above): keeping `Wire`. The case for `JSON` is reasonable but it's a coin-flip semantic call, not a clear win. Reasoning:

- The Date/Buffer analogy in the changeset is for the type+const colocation pattern, not specifically the `toJSON`/`fromJSON` method naming. `Date.toJSON()` is an instance method called by `JSON.stringify` auto-invocation; our static factory pattern is closer to Uniswap's `CurrencyAmount.fromRawAmount` (which doesn't use `JSON` naming) than to `Date.toJSON`.
- `Wire` is the established term in the web3/SDK ecosystem (viem `encodeAbiParameters`, Protobuf "wire format", ABI encoding). Renaming would diverge from that convention.
- All the design docs (`SDK_PAYMENTINFO_NAMESPACE.md`, `SDK_PAYMENTINFO_ERGONOMICS_DECISION.md`) use `Wire` throughout — code/doc drift cost.
- This would be the third rename in this PR's lifetime (`toPaymentInfo` → `reconstructPaymentInfoStruct` → `reconstructPaymentInfoWire`). Marginal semantic gain doesn't justify the churn.
- `Wire` survives future format additions (e.g., if a non-JSON serialization is ever supported); `JSON` would be misleading then.

Logged for future re-evaluation if `JSON` naming becomes more widely conventional in the ecosystem.

A1igator:
### SDK Review — Round 3

Found 0 new issues since ff489af (axes: tests, conventions, dead code, example alignment, SDK design).

**Resolved since round 2:**
- Stale `PaymentInfoStruct` JSDoc reference on `PaymentInfo.fromWire` (`packages/core/src/types/index.ts:49`). JSDoc now reads "Convert a JSON-form `PaymentInfoWire`" cleanly, with the parenthetical removed.
- Test comment in `packages/helpers/tests/forward-to-arbiter.test.ts:96` updated to "the reconstructed PaymentInfoWire", matching the asserted field name.
- `salt` and `maxAmount` validator error messages and JSDoc now align with runtime behavior (`expected uint256 (decimal or 0x-prefixed hex)`), which is what `BigInt(...)` actually accepts. Resolves the doc-runtime mismatch carried from round 1 without changing acceptance semantics.

**Still open:** none.
**New:** none.

---

A1igator:
### Naming follow-up: `PaymentInfoJSONSafe` (non-blocking)

Fair point on the `toJSON`/`fromJSON` collision concern. The `Object.prototype.toJSON` auto-call convention pattern-matches to that name even though the static-namespace usage here wouldn't actually trigger it.

One alternative worth considering: **`PaymentInfoJSONSafe`** with `toJSONSafe` / `fromJSONSafe`.

- Precisely names what the form is: the JSON-stringifiable shape of `PaymentInfo`, no more, no less.
- No collision with `Object.prototype.toJSON` (the literal name `toJSON` is what `JSON.stringify` auto-calls; `toJSONSafe` doesn't).
- Doesn't carry the "wire = binary" connotation that `Wire` has in the broader web3 ecosystem (viem `encodeAbiParameters`, ABI encoding, Protobuf wire format all refer to byte-level serialization).
- Survives the "what makes this form distinct from `PaymentInfo`" test: the only meaningful difference is JSON-safety (bigints become strings).

Cost: verbose. Five characters longer than `Wire`, two extra in every consumer call.

Other alternatives considered and ruled out:
- `Raw`: `PaymentInfo` is already the canonical ABI-derived form (`AbiParameterToPrimitiveType<AuthorizeFn['inputs'][0]>`); the string-form is a derivative encoding, not "more raw."
- `Serialized`: implies string output, but the form is an object.
- `Encoded`: implies bytes (viem convention).
- `serialize`/`deserialize`: overkill for `bigint.toString()` on two fields.

If keeping `Wire` and the verbosity of `JSONSafe` isn't worth it, that's a defensible call. Last suggestion on this thread.

---

A1igator (approved):

vraspar:
Thanks — keeping `Wire` for the same reasons as the prior naming reply, plus:

- `JSONSafe` is a novel term (not idiomatic in JS/TS or web3 SDKs); inventing vocabulary to escape both incumbents is a smell.
- Verbosity: `PaymentInfo.toJSONSafe()` reads worse than `.toWire()` and adds characters to every call site.
- We've already cycled through three names this PR — third rename for a marginal semantic gain isn't worth the churn.
- "Wire = binary in web3" is a real but mild concern; the design docs + PR description establish what `Wire` means in this codebase.

Appreciate the thorough review. Treating this thread as closed.

PR #127, opened 2026-05-18
CI: fold scenario:dispute + scenario:permit2-charge into test-fork

vraspar:
## Summary

PR 1 of 3 in the `@x402r/sdk@0.3.0` publish-readiness series tracked at `x402r-notes/plans/PHASE_3_PUBLISH_READINESS.md`. Extends the existing `test-fork` CI job to run two end-to-end scenarios after the fork test suite, and fixes three misleading "HTTP 402 flow" labels in scenario output.

### Why these two scenarios

**`scenario:permit2-charge`** — we have zero on-chain Permit2 coverage today.

```
$ grep -r permit2 packages/core/tests/integration/
(no matches)
```

The only existing Permit2 test is `packages/core/tests/permit2.test.ts`, a unit test with mocked viem. The permit2-charge scenario runs the full payer-side flow against a real ERC-20 on an Anvil fork: `ERC20.approve(PERMIT2, MAX)` then `signPermit2Authorization` then atomic `payment.charge`, asserting real balance deltas. Without this in CI, a Permit2 regression ships silently.

**`scenario:dispute`** — the only multi-step lifecycle test we have. The 7-step flow exercises cross-call state mutations (authorize → refund request → evidence submission by payer + merchant → arbiter review → `voidPayment` → hook auto-approval → fee distribution) that per-action fork tests cannot fake. Each per-action fork test starts from a fresh fork; only an end-to-end scenario validates that state from step N is correctly consumed by step N+1.

### Why not the other scenarios

`scenario:capture`, `scenario:atomic-charge`, and `scenario:partial-refund-flow` are deliberately excluded — they're redundant with per-action fork tests in `packages/core/tests/integration/` (`authorize-capture.fork.test.ts`, `direct-charge.fork.test.ts`, `partial-refund.fork.test.ts`). Running them in CI would burn wall-clock without adding coverage.

### CI shape

The existing `test-fork` job is extended in place — no new CI lane added. Wall-clock delta is ~8 seconds (both scenarios complete in <2s each on a warm fork; chain RPC dominates).

Concretely:
- New `Run CI scenarios (dispute + permit2)` step appended after `Run fork tests`, using the same `nick-fields/retry@v3` wrapper with `timeout_minutes: 5, max_attempts: 2` to absorb public Base Sepolia RPC flake.
- Job `timeout-minutes` bumped `15 → 20` to give retry headroom.
- New root script `pnpm scenarios:ci` chains the two scenarios.

### Label fix (commit 1)

Three scenario steps printed `"Authorize payment via HTTP 402 flow"` while doing direct SDK calls via `signReceiveAuthorization` + `payment.authorize` — no HTTP, no 402, no facilitator, no x402 wire format. Renamed to `"Authorize payment via SDK viem flow"` in `happy-path-capture.ts`, `dispute-resolution.ts`, and `partial-refund-flow.ts` so scenario output matches reality.

## Test plan

- [x] `pnpm scenario:dispute` passes locally (7 steps, ~1.8s)
- [x] `pnpm --filter examples scenario:permit2-charge` passes locally (3 steps, ~2s)
- [x] `pnpm scenarios:ci` chain works end-to-end (first run failed on anvil port reuse, retry succeeded — exactly the case `nick-fields/retry@v3` is configured to handle)
- [ ] CI `test-fork` job green on this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,12 +109,12 @@ jobs:
       - name: Run CI scenarios (dispute + permit2)
         uses: nick-fields/retry@v3
         with:
+          # Retry guards against public Base Sepolia RPC flake on the upstream
+          # fork. Port-collision band-aid is no longer needed: the orchestrator
+          # at examples/scripts/scenarios-ci.ts runs ONE prool with per-scenario
+          # URL subpaths, so retries don't race a TIME_WAIT port.
           timeout_minutes: 5
           max_attempts: 2
-          # Wait between attempts to avoid anvil port-reuse collisions.
-          # TODO: derive ANVIL_PORT from VITEST_POOL_ID / process.pid in
-          # examples/shared/anvil-setup.ts so retries get a fresh port.
-          retry_wait_seconds: 10
           command: pnpm scenarios:ci
 
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
     # Fork tests do not contribute to code coverage (no --coverage flag).
     # They validate on-chain integration via Anvil forks.
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v4
       - name: Install Foundry
@@ -106,6 +106,12 @@ jobs:
           timeout_minutes: 10
           max_attempts: 2
           command: pnpm test:fork
+      - name: Run CI scenarios (dispute + permit2)
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 5
+          max_attempts: 2
+          command: pnpm scenarios:ci
 
   build:
     name: Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
     # Fork tests do not contribute to code coverage (no --coverage flag).
     # They validate on-chain integration via Anvil forks.
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
       - name: Install Foundry
@@ -111,6 +111,10 @@ jobs:
         with:
           timeout_minutes: 5
           max_attempts: 2
+          # Wait between attempts to avoid anvil port-reuse collisions.
+          # TODO: derive ANVIL_PORT from VITEST_POOL_ID / process.pid in
+          # examples/shared/anvil-setup.ts so retries get a fresh port.
+          retry_wait_seconds: 10
           command: pnpm scenarios:ci
 
   build:
--- a/examples/scenarios/README.md
+++ b/examples/scenarios/README.md
@@ -27,7 +27,7 @@ Exercises the complete dispute flow:
 4. Both parties submit evidence
 5. Arbiter reviews evidence and approves refund
 6. Verify refund amounts
-7. Distribute protocol fees
+7. Verify zero protocol fees accrued
 
 ### atomic-charge
 
--- a/examples/scenarios/README.md
+++ b/examples/scenarios/README.md
@@ -22,7 +22,7 @@ Demonstrates the simplest payment lifecycle — merchant authorizes a payment, w
 
 Exercises the complete dispute flow:
 1. Deploy marketplace operator
-2. Authorize payment (HTTP 402 flow)
+2. Authorize payment via SDK viem flow
 3. Payer requests refund
 4. Both parties submit evidence
 5. Arbiter reviews evidence and approves refund
--- a/examples/scenarios/dispute-resolution.ts
+++ b/examples/scenarios/dispute-resolution.ts
@@ -35,9 +35,9 @@ async function main() {
     }
 
     // ================================================================
-    // Step 1: Authorize payment (HTTP 402 flow)
+    // Step 1: Authorize payment (direct SDK call)
     // ================================================================
-    runner.step('Authorize payment via HTTP 402 flow')
+    runner.step('Authorize payment via SDK viem flow')
 
     const payerAccount = privateKeyToAccount(PAYER_PRIVATE_KEY)
     const { collectorData, tokenCollector } = await signReceiveAuthorization({
--- a/examples/scenarios/dispute-resolution.ts
+++ b/examples/scenarios/dispute-resolution.ts
@@ -11,7 +11,7 @@ import { StepRunner } from './runner.js'
 // Flow: authorize → payer requests refund → payer + merchant submit evidence →
 //       arbiter reviews evidence → merchant executes voidPayment
 //       (hook approves automatically) → verify refund amounts →
-//       merchant distributes fees
+//       verify zero protocol fees accrue
 // ---------------------------------------------------------------------------
 
 async function main() {
@@ -191,9 +191,9 @@ async function main() {
     )
 
     // ================================================================
-    // Step 7: Distribute fees
+    // Step 7: Verify zero protocol fees accrued
     // ================================================================
-    runner.step('Merchant distributes protocol fees')
+    runner.step('Verify zero protocol fees accrued')
 
     const accumulatedFees =
       await ctx.merchant.operator.getAccumulatedProtocolFees(
--- a/examples/scenarios/dispute-resolution.ts
+++ b/examples/scenarios/dispute-resolution.ts
@@ -52,6 +52,15 @@ async function main() {
     // authorize collected zero tokens, before === after and delta === 0
     // either way.
     const payerBalanceBefore = await readBalance(ctx.accounts.payer)
+    // Snapshot receiver + feeReceiver too — under a full refund neither party
+    // should see a balance change. A non-zero delta on either would indicate
+    // tokens were misdirected (e.g. a future refactor accidentally routes a
+    // partial-fee path through a full-void). Catches regressions the
+    // payer-only snapshot would miss.
+    const receiverBalanceBefore = await readBalance(ctx.accounts.merchant)
+    const feeReceiverBalanceBefore = await readBalance(
+      ctx.paymentInfo.feeReceiver,
+    )
 
     // ================================================================
     // Step 1: Authorize payment (direct SDK call)
@@ -192,6 +201,23 @@ async function main() {
     runner.assert(
       payerDelta === 0n,
       `payer balance net-zero after full refund (authorized → refunded back); actual delta ${payerDelta}`,
+    )
+
+    // Receiver + feeReceiver must not have moved at all under a full refund.
+    // Distinct from the payer net-zero check above: payer can net-zero even if
+    // tokens were misrouted (e.g. paid out and then refunded back). These
+    // deltas pin the destinations explicitly.
+    const receiverBalanceAfter = await readBalance(ctx.accounts.merchant)
+    const feeReceiverBalanceAfter = await readBalance(
+      ctx.paymentInfo.feeReceiver,
+    )
+    runner.assert(
+      receiverBalanceAfter === receiverBalanceBefore,
+      `receiver balance unchanged under full refund; actual delta ${receiverBalanceAfter - receiverBalanceBefore}`,
+    )
+    runner.assert(
+      feeReceiverBalanceAfter === feeReceiverBalanceBefore,
+      `feeReceiver balance unchanged under full refund; actual delta ${feeReceiverBalanceAfter - feeReceiverBalanceBefore}`,
     )
 
     // ================================================================
--- a/examples/scenarios/dispute-resolution.ts
+++ b/examples/scenarios/dispute-resolution.ts
@@ -46,7 +46,11 @@ async function main() {
     // refund actually moved tokens. Under a full refund, the payer pays
     // PAYMENT_AMOUNT on authorize and receives PAYMENT_AMOUNT on refund, so
     // the net delta must be exactly 0. A non-zero delta means tokens never
-    // moved back (silent refund failure).
+    // moved back (silent refund failure). The `hasCollectedPayment` assertion
+    // at Step 1 (after authorize) is the load-bearing gate that prevents this
+    // delta check from vacuously passing on a silent authorize failure — if
+    // authorize collected zero tokens, before === after and delta === 0
+    // either way.
     const payerBalanceBefore = await readBalance(ctx.accounts.payer)
 
     // ================================================================
--- a/examples/scenarios/dispute-resolution.ts
+++ b/examples/scenarios/dispute-resolution.ts
@@ -34,6 +34,21 @@ async function main() {
       throw new Error('Evidence module not available')
     }
 
+    const readBalance = (owner: `0x${string}`) =>
+      ctx.publicClient.readContract({
+        address: ctx.paymentInfo.token,
+        abi: erc20Abi,
+        functionName: 'balanceOf',
+        args: [owner],
+      })
+
+    // Snapshot the payer's funded post-setup balance so Step 6 can assert the
+    // refund actually moved tokens. Under a full refund, the payer pays
+    // PAYMENT_AMOUNT on authorize and receives PAYMENT_AMOUNT on refund, so
+    // the net delta must be exactly 0. A non-zero delta means tokens never
+    // moved back (silent refund failure).
+    const payerBalanceBefore = await readBalance(ctx.accounts.payer)
+
     // ================================================================
     // Step 1: Authorize payment (direct SDK call)
     // ================================================================
@@ -162,13 +177,18 @@ async function main() {
       'Refundable amount === 0 after refund executed',
     )
 
-    const payerUsdcBalance = await ctx.publicClient.readContract({
-      address: ctx.paymentInfo.token,
-      abi: erc20Abi,
-      functionName: 'balanceOf',
-      args: [ctx.accounts.payer],
-    })
+    const payerUsdcBalance = await readBalance(ctx.accounts.payer)
     runner.log(`Payer USDC balance: ${payerUsdcBalance}`)
+
+    // Net-zero invariant: payer paid PAYMENT_AMOUNT on authorize and got
+    // PAYMENT_AMOUNT back on full refund. A delta of -PAYMENT_AMOUNT would
+    // indicate the refund silently failed to move tokens.
+    const payerBalanceAfter = await readBalance(ctx.accounts.payer)
+    const payerDelta = payerBalanceAfter - payerBalanceBefore
+    runner.assert(
+      payerDelta === 0n,
+      `payer balance net-zero after full refund (authorized → refunded back); actual delta ${payerDelta}`,
+    )
 
     // ================================================================
     // Step 7: Distribute fees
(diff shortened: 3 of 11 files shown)


vraspar:
### SDK Review

Found 4 issues (axes: tests, ci config):

1. **[Tests]** `scenario:dispute` Step 6 reads payer balance, logs it, never asserts it. The whole point of dispute resolution is "refund returned the funds to the payer." Step 6 reads `payerUsdcBalance`, prints it, moves on — the scenario passes whether the refund moved tokens or not. Folding this into CI as-is adds a green check without the coverage the PR description claims. Fix: snapshot `payerBalanceBefore` at Step 1 and assert `payerBalanceAfter - payerBalanceBefore === PAYMENT_AMOUNT` at Step 6.

   https://github.com/BackTrackCo/x402r-sdk/blob/0f74515f3edd2b31b5cb6deff8904a841a446eca/examples/scenarios/dispute-resolution.ts#L165-L171

2. **[Tests]** `scenario:dispute` Step 7 fee-distribution branch is unreachable in the CI'd path. A full-refund dispute produces `accumulatedFees === 0n`, so the `if (accumulatedFees > 0n)` branch never executes and the `else` only logs. The only assertion in Step 7 (`remainingFees === 0n`) never fires under the dispute path. Either assert `accumulatedFees === 0n` explicitly (pinning that a full refund collects zero protocol fees) or move the fee-distribution exercise to a scenario where fees actually accrue.

   https://github.com/BackTrackCo/x402r-sdk/blob/0f74515f3edd2b31b5cb6deff8904a841a446eca/examples/scenarios/dispute-resolution.ts#L184-L197

3. **[CI config]** Job timeout (20m) is below worst-case retry budget (~30m). `test:fork` at `timeout_minutes: 10, max_attempts: 2` = up to 20 min. The new `scenarios:ci` step at `timeout_minutes: 5, max_attempts: 2` = up to 10 min. Worst case 30 min vs `timeout-minutes: 20` job ceiling. The job gets killed mid-retry under pathological flake. Either bump the job timeout to 30 min or drop the scenario step's `timeout_minutes` to 3.

   https://github.com/BackTrackCo/x402r-sdk/blob/0f74515f3edd2b31b5cb6deff8904a841a446eca/.github/workflows/ci.yml#L84-L112

4. **[CI config]** Retry masks a known port-reuse flake instead of fixing it. The commit message says "first run failed on anvil port reuse, retry succeeded." Without `retry_wait_seconds`, the second attempt may collide on the same port again — the green is by luck, not design. Add `retry_wait_seconds: 10` to the retry block, or root-cause the port collision in the scenario runner.

   https://github.com/BackTrackCo/x402r-sdk/blob/0f74515f3edd2b31b5cb6deff8904a841a446eca/.github/workflows/ci.yml#L107-L112

`scenario:permit2-charge` has strict balance-delta assertions and no slop — clean. The label fix (`"HTTP 402 flow"` → `"SDK viem flow"`) is complete across all 3 scenarios + the README. Minor follow-up worth filing separately: `permit2-charge` is not listed in `examples/scenarios/README.md`.

---

vraspar:
All 4 review findings addressed:

1. **Dispute Step 6 — payer balance delta assertion** (commit d7cc449): snapshotted `payerBalanceBefore` after `setup({ authorize: false })` and asserted `payerDelta === 0n` at Step 6. Net-zero is correct here because the payer pays `PAYMENT_AMOUNT` on authorize and receives `PAYMENT_AMOUNT` back on full refund; a non-zero delta now fires a real assertion failure instead of being silently logged.
2. **Dispute Step 7 — explicit `accumulatedFees === 0n`** (commit d7cc449): the previous `if (accumulatedFees > 0n)` branch was dead under a full refund, so the only assertion in Step 7 never ran. Replaced with an unconditional `accumulatedFees === 0n` assertion and dropped the dead conditional.
3. **`test-fork` timeout bump 20 → 30m** (commit 8a60f5c): worst-case retry budget is `test:fork` 10×2 + `scenarios:ci` 5×2 = 30 min, so the previous 20-min cap could time out mid-retry.
4. **`retry_wait_seconds: 10` on the scenarios step** (commit 8a60f5c): without a wait between attempts, the retry could re-collide on the same hardcoded `ANVIL_PORT = 8846`.

Verified locally: `pnpm scenario:dispute` passes with both new assertions; `pnpm scenarios:ci` (dispute + permit2) passes end-to-end.

**Follow-up filed (not in this PR):** the real fix for the port-reuse flake is to derive `ANVIL_PORT` from `VITEST_POOL_ID` (or `process.pid` for non-vitest tsx runs) in `examples/shared/anvil-setup.ts:76` so each retry/worker gets a fresh port. Left a `TODO` next to `retry_wait_seconds` pointing to that file. Will open as a separate PR.

vraspar:
Round 3 follow-up — addressed the MAJOR finding on `examples/scenarios/permit2-charge.ts` step 1's brittle `after > before` assertion in 4fe3a10.

**Chosen fix (minimal robust shape):** assert only the post-condition the scheme actually needs, `after >= PAYMENT_AMOUNT`. This tolerates either fork state — zero starting allowance or upstream-inherited MAX — because both satisfy the invariant the upcoming `payment.charge` requires.

**Why not the precondition variant** (`before < PAYMENT_AMOUNT` then `after > before`)? It still fails loudly under the exact scenario the review flagged: if the unpinned Base Sepolia fork inherits a prior MAX_UINT256 approval for the anvil #1 payer key (a well-recycled test key), the precondition itself trips and the scenario reports a misleading failure even though sufficient allowance is in place. That's a rebrand of the error, not a fix.

Also softened the docstring's "fork starts with zero allowance" claim to reflect that step 1 is idempotent and the post-condition pins the real invariant.

Verified: `pnpm --filter examples scenario:permit2-charge` and `pnpm scenarios:ci` both pass.

A1igator:
### SDK Review — Round 2

Found 1 new issue since commit 0f74515 (axes: tests, ci config).

- Step 6 payer-balance-delta assertion added (commit d7cc449).
- Step 7 explicit `accumulatedFees === 0n` assertion replacing dead `if` branch (d7cc449).
- `test-fork` job `timeout-minutes` bumped to 30 (8a60f5c).
- `retry_wait_seconds: 10` added to scenarios step with TODO for proper port-fix (8a60f5c).

**Still open:** none.

**New:**

1. **[Tests]** Step 7 label/comments still describe an action the code no longer performs. The fix for the prior unreachable-branch issue dropped the `distributeFees()` call entirely (correct — a full refund collects zero fees), but the surrounding framing wasn't updated: `runner.step('Merchant distributes protocol fees')`, the `// Step 7: Distribute fees` comment, the file-header flow comment `merchant distributes fees`, and `README.md` step 7 `Distribute protocol fees`. The scenario now asserts "no fees accrue," not "fees are distributed." This is the exact class of label-vs-reality mismatch commit 1 of this PR was fixing. Rename the step to something like `Verify zero protocol fees accrued` and update the four surrounding references.

   https://github.com/BackTrackCo/x402r-sdk/blob/4fe3a10736a700b6817990456237164f646a8432/examples/scenarios/dispute-resolution.ts#L194-L210

---

vraspar:
Round 2 A1igator finding addressed (52eafea).

The teammate-round Step 7 fix replaced the dead distribute branch with an `accumulatedFees === 0n` assertion, but the surrounding labels still described the old behavior. Renamed all four references to **"Verify zero protocol fees accrued"** so the docs match the assertion:

- `examples/scenarios/dispute-resolution.ts:14` — flow comment ("merchant distributes fees" → "verify zero protocol fees accrue")
- `examples/scenarios/dispute-resolution.ts:194` — Step 7 section comment
- `examples/scenarios/dispute-resolution.ts:196` — `runner.step(...)` label
- `examples/scenarios/README.md:30` — step 7 entry

Honest miss: this is the exact same label-vs-reality anti-pattern the first commit of this PR fixed for HTTP 402. Should have caught it when applying the Step 7 logic fix. `pnpm scenarios:ci` still passes.

A1igator:
### SDK Review — Round 3

No new or open issues since commit 4fe3a10.

- Step 7 label/comments + README renamed to `Verify zero protocol fees accrued` (52eafea).

**Still open:** none.
**New:** none.

---

A1igator (dismissed):

vraspar:
Follow-up notes from a deeper read (filing here so subsequent PRs can pick them up):

**`ANVIL_PORT` TODO is the proper fix, not a workaround.** Fork tests don't hit the port-reuse collision because they share one Anvil on port 8745 and isolate vitest workers via per-`VITEST_POOL_ID` RPC subpaths (`packages/core/tests/setup/anvil.ts:43,106-118`). Scenarios spawn an independent Anvil per `setup()` call on hardcoded port 8846 (`examples/shared/anvil-setup.ts:76`), so retries can EADDRINUSE without the 10s wait. Deriving the port from `VITEST_POOL_ID`/`process.pid` removes the wait + the source of flake.

**Job timeout 15→30 reflects retry budget, not the +8s happy-path delta.** Worst case = `10min × 2` (fork tests) + `5min × 2` (scenarios) = 30 min.

**Permit2 assertion change is correct, but for a different reason than the inline comment claims.** The "warm-fork inherits MAX upstream" scenario can't actually happen — the payer is Anvil deterministic account #1, has never interacted with Permit2 on Base Sepolia upstream, and each scenario starts a fresh Anvil (`examples/shared/anvil-setup.ts:99`) with no state carryover. The real justification is that `after >= PAYMENT_AMOUNT` asserts the contract invariant the upstream `charge` needs (allowance covers payment) rather than an implementation detail of `createPermit2ApprovalTx` (`packages/core/src/payment/permit2.ts:165` hard-codes MAX). Robust against future helper changes either way — comment can be tightened in a followup.

**Dispute net-zero invariant relies on the earlier `hasCollectedPayment === true` check (`dispute-resolution.ts:72-79`).** Standalone, `payerDelta === 0n` would pass vacuously if `authorize()` silently collected zero tokens. Safe in context but worth noting if anyone reorders steps.

vraspar:
Two follow-ups landed on this branch:

**Restored e75c1af (comment fixes).** The previous revert (3e232ec) was a mistake — un-reverted it via git revert to bring back the Permit2 Step 1 contract-invariant rationale and the dispute Step 6 net-zero note. New commit: 980d24e.

**Shared-prool architecture for scenarios:ci.** New commit: 1994fe8. Root cause of the EADDRINUSE flake the retry_wait_seconds band-aid was masking: each scenario was spawning its own prool on port 8846, so rapid sequential or retried runs raced TIME_WAIT.

Refactored to mirror how fork tests already share prool (packages/core/tests/setup/anvil.ts:43, packages/core/tests/setup/globalSetup.ts:9):

- examples/scripts/scenarios-ci.ts: ONE prool server per CI run; each scenario subprocess gets SCENARIO_RPC_URL pointing at its own numeric pool key (/1, /2). prool routes each key to its own forked Anvil child.
- examples/shared/anvil-setup.ts: when SCENARIO_RPC_URL is set, skip the Server.create block and use the orchestrator's URL; cleanup becomes a no-op. Unset = unchanged standalone behavior (`pnpm scenario:capture` etc. still work).
- examples/scenarios/permit2-charge.ts: replaced a hardcoded http://127.0.0.1:8846/1 transport with ctx.rpcUrl (the old URL routed permit2's approval tx to dispute's Anvil child, where it never mined).
- ci.yml: dropped retry_wait_seconds: 10 and the TODO at anvil-setup.ts:76. max_attempts: 2 and timeout_minutes: 5 stay — they still guard against public Base Sepolia RPC flake.

Side benefit: ONE upstream fork operation per CI run instead of one per scenario — measurable wallclock improvement on slow public RPC.

Verified locally: standalone `scenario:dispute` + `scenario:permit2-charge` still pass; `scenarios:ci && scenarios:ci` back-to-back passes (no EADDRINUSE).

vraspar:
### SDK Review — Round 4

No new or open issues since commit 4fe3a10.

- Permit2 Step 1 comment rewritten with the contract-invariant rationale (980d24e — reapply of e75c1af after an inadvertent revert).
- Dispute Step 6 comment notes the `hasCollectedPayment` gate is what prevents the net-zero delta check from vacuously passing on silent authorize failure (980d24e).
- `scenarios:ci` now runs ONE prool with per-scenario URL subpaths (`examples/scripts/scenarios-ci.ts`), replacing the chained-pnpm + `retry_wait_seconds: 10` band-aid. Mirrors the fork-test pattern at `packages/core/tests/setup/anvil.ts:43` (1994fe8).
- `permit2-charge.ts:38` hardcoded `http://127.0.0.1:8846/1` replaced with `ctx.rpcUrl` — fixes a latent bug where permit2's approval tx would have routed to dispute's Anvil child under shared-prool mode (1994fe8).
- CI yml `retry_wait_seconds: 10` and the corresponding TODO removed; `max_attempts: 2` + `timeout_minutes: 5` retained as public-RPC flake guard (1994fe8).

**Still open:** none.
**New:** none.

---

vraspar:
Folded in the round-6 reviewer's below-threshold note (F) as defense-in-depth: 02082cc.

The previous `SCENARIOS` array hardcoded `{ file, key }` pairs, so a future copy-paste mistake giving two scenarios the same numeric key would have routed both subprocesses to the same Anvil child via prool — silent state crossover with no validation to catch it. Replaced with a flat `SCENARIO_FILES` array; the key is now derived as `index + 1` in the loop, so duplicates are impossible by construction.

Verified locally with `pnpm scenarios:ci && pnpm scenarios:ci` back-to-back — both runs pass.

A1igator:
### SDK Review — Round 4

No new or open issues since commit 52eafea.

- `retry_wait_seconds` band-aid replaced with proper root-cause fix: orchestrator at `examples/scripts/scenarios-ci.ts` runs one prool with per-scenario subpath keys, mirroring `packages/core/tests/setup/anvil.ts:43` (1994fe8).
- `examples/scripts/scenarios-ci.ts` derives the pool key from array index instead of hardcoded `{ file, key }` pairs — duplicate-key state crossover impossible by construction (02082cc).
- Permit2 + dispute scenario rationale comments rewritten to reflect the actually-load-bearing reasoning (980d24e).

**Still open:** none.
**New:** none.

---

A1igator (dismissed):

vraspar:
Fresh review pass — one merge blocker + a few polish items.

**Blocker**

- `examples/scripts/scenarios-ci.ts:79-82` — no `SIGINT`/`SIGTERM`/`unhandledRejection` handlers. Ctrl+C or CI cancellation mid-run leaks the prool server (port stays bound, blocks the next run). Add `process.on('SIGINT'/'SIGTERM', async () => { await server.stop(); process.exit(130) })` plus an `unhandledRejection` mirror that calls `server.stop()` before exiting.

**Correctness — medium**

- `examples/scripts/scenarios-ci.ts:42-46` — `child.on('error')` (e.g., `tsx` missing → ENOENT) and `child.on('exit', code)` both resolve the same promise, so spawn failures can't be distinguished from clean exits. Reject on `error` or surface the spawn error explicitly.
- `examples/scripts/scenarios-ci.ts:65-72` — empty `SCENARIO_FILES` makes the loop a no-op and exits 0 silently. Guard with `if (SCENARIO_FILES.length === 0) throw new Error('no scenarios')`.
- `examples/shared/anvil-setup.ts:108-111` — malformed `SCENARIO_RPC_URL` isn't validated up front; failure surfaces later as a cryptic viem RPC error. Wrap with `new URL(sharedRpcUrl)` at entry for fail-fast.

**Correctness — low**

- `examples/scripts/scenarios-ci.ts:61` — `await server.start()` is assumed to block until prool is listening, but no comment documents that contract. One-line comment.

**Test strengthening (nice to have)**

- `examples/scenarios/permit2-charge.ts:54-56` — the new `after >= PAYMENT_AMOUNT` is correct as the contract invariant, but a regression in the approval helper that sets allowance to 0 would now pass silently. `after >= PAYMENT_AMOUNT && after > before` catches both classes.
- `examples/scenarios/dispute-resolution.ts:25-53` — net-zero invariant only snapshots payer balance. If a future bug misdirected refund tokens to receiver or feeReceiver, this wouldn't catch it. Snapshot all three for full coverage.

**Convention nit**

- `examples/scripts/` is a new directory — sibling CLI scripts live one level up (`examples/payer/*.ts`, `examples/merchant/*.ts`). Defensible because `scenarios-ci.ts` is infrastructure (not a role-based scenario), but worth a one-line README note if it stays.

The shared-prool + per-scenario-subpath architecture itself is sound. Findings are all on the error/cancellation-handling paths or polish, not the core design.

vraspar:
Round-5 fixes pushed. Summary:

**Applied:**

- **Blocker — signal handlers** (`2884d0a`): SIGINT/SIGTERM/unhandledRejection all stop prool before exit. Verified by hand: SIGINT during a run prints `[scenarios-ci] received SIGTERM, stopping prool…` and port 8846 frees within ~5s. Without these, cancellation left anvil + the listener bound and the next run got EADDRINUSE.
- **Spawn-error vs exit-code distinction** (`1470a55`): `runScenario` now rejects on `child.on('error')` so ENOENT-class failures surface as "failed to spawn …" instead of being masked as "scenario exited 1." The outer loop wraps spawn errors in try/catch so `finally { await server.stop() }` still runs.
- **Misconfiguration guards** (`d4db63b`): empty `SCENARIO_FILES` throws (no silent no-op CI step); `SCENARIO_RPC_URL` is `new URL()`-validated up front so a malformed env var fails with a clear message rather than mid-test through viem.
- **`server.start()` blocking-contract comment** (`1b7060c`): one-liner above `await server.start()`.
- **Dispute multi-balance snapshot** (`5af9b04`): receiver + feeReceiver balances now snapshot pre-refund and asserted unchanged post-refund. The payer net-zero check would have missed a misroute (paid out + refunded back nets the same delta); the destination snapshots pin it explicitly.
- **`examples/scripts/README.md`** (`06e6bb8`): infrastructure-vs-role-based scripts placement note.

Verification: `pnpm scenario:dispute` standalone passes (new assertions hold), `pnpm scenarios:ci` passes, back-to-back invocations pass (no EADDRINUSE), and SIGINT mid-run shuts prool down cleanly.

**Pushback — Permit2 dual-assertion:**

Not making this change in this PR — surfacing for discussion.

The suggested `after >= PAYMENT_AMOUNT && after > before` reintroduces the brittleness the round-3 MAJOR was originally fixing. If a warm fork inherits `before === MAX` and the approval helper sets MAX again (current helper behavior), then `after === before === MAX`, so `after > before` is false and the test fails — exactly the symptom we just removed.

The split form below preserves both invariants without that conflict:

```ts
runner.assert(after >= PAYMENT_AMOUNT, ...)
if (before < PAYMENT_AMOUNT) {
  runner.assert(after > before, 'approval tx actually wrote state (helper not silently broken)')
}
```

The `if` guard ensures the progress check fires only when it would matter (when `before` was insufficient and the helper had to do work); when `before` was already sufficient, no-op is correct behavior. Happy to land this as a follow-up if you'd like it.

A1igator:
### SDK Review — Round 5

No new or open issues since commit 02082cc.

- SIGINT/SIGTERM/unhandledRejection handlers prevent leaked anvil child + bound port on cancellation (2884d0a).
- `runScenario` rejects on spawn-time errors instead of conflating ENOENT-class failures with exit-code-1 crashes (1470a55).
- Dispute Step 6 now snapshots receiver + feeReceiver and asserts both deltas are zero — catches misrouted-token regressions the payer-only net-zero check would miss (5af9b04).
- `SCENARIO_FILES` empty-list guard and `SCENARIO_RPC_URL` URL validation added (d4db63b).
- `SCENARIO_FILES` typed as `readonly string[]` so the empty-list guard isn't dead code under strict TS (1438b7f).
- `examples/scripts/README.md` documents the infrastructure-vs-role-example distinction (06e6bb8).
- `Server.start()` blocking-until-listening contract documented inline (1b7060c).

**Still open:** none.
**New:** none.

---

A1igator (approved):

PR #128, opened 2026-05-18
core: partial-capture edge case tests (void-after-full, double-application)

vraspar:
PR 2 of 3 in the `@x402r/sdk@0.3.0` publish-readiness series tracked at `x402r-notes/plans/PHASE_3_PUBLISH_READINESS.md`.

## Summary

Adds two new `describe` blocks to `packages/core/tests/integration/partial-refund.fork.test.ts` to pin down on-chain behavior at the partial-capture / void boundary. No new files, no new fixtures, no contract changes.

### (b) `voidPayment` after full capture

**Real-world failure mode:** a merchant calls `capture(full)` to claim the full authorized amount, then mistakenly (or defensively) calls `voidPayment()` afterward. The JSDoc on `voidPayment` (`packages/core/src/actions/refund-budget/voidPayment.ts:17-18`) says it "empties the authorization in one transaction" but does not specify what happens when `capturableAmount === 0` already.

**On-chain surprise class:** a future contract change could turn the current revert into a silent no-op success, or vice versa. Either direction would change the SDK's effective surface without an SDK-side code change.

**Observed behavior (pinned 2026-05-18 against the fixture-deployed `PaymentOperator` + `EscrowPeriod`):**
- The SDK's `voidPayment()` submits the tx and returns a tx hash WITHOUT throwing. This is because `walletClient.writeContract` on anvil does not simulate before sending, and the SDK's `wrapContractCall` error-wrapping only fires on thrown errors.
- The on-chain `void()` REVERTS in the receipt (`receipt.status === 'reverted'`).
- `capturableAmount` remains `0n` (no state corruption).

The test asserts the receipt-level revert + post-state, which is the contract-level guarantee that matters.

### (c) Double-application / replay protection

**Real-world failure mode:** an integration retries or re-applies a `paymentInfo + signature` after the original authorization was already `capture+void`'d. If the escrow nonce ever stopped rejecting this, an attacker (or a buggy retry loop) could re-collect tokens from a payer who thought their payment was fully consumed.

**On-chain surprise class:** the security boundary here is `computeEscrowNonce` at `packages/core/src/payment/hashing.ts:87-97` — derived from `(chainId, escrowAddress, paymentInfo with payer = address(0))`. ERC-3009 nonces are single-use, so a second `authorize` on the same `paymentInfo` (same salt) must fail.

**Observed behavior (pinned 2026-05-18):**
- Same shape as case (b): the SDK's `authorize()` returns a tx hash without throwing, but the on-chain tx REVERTS in the receipt.
- `capturableAmount` stays `0n` after the failed replay attempt, confirming nothing was re-collected.

The test re-uses the file-level `beforeAll` `paymentInfo` (`salt: 3n`) which the preceding `Scenario 3: Partial capture + void remainder` describe block has already consumed. Vitest runs describes in file order, so this dependency is well-defined; the test has a code comment documenting it.

## Out of scope

Edge case (a) — `merchantAmount === 0` — is deferred. It needs contracts-side investigation in `x402r-contracts` (whether the operator allows a zero-amount capture at all, and what the intended behavior is) before an SDK-level test can meaningfully pin it down.

## CI cost delta

~30s on the existing `test-fork` job (two new tests, each bounded at 60s but in practice each completes in ~2-3s once fixtures are warm). No new files, no new fixtures, no new CI lanes.

## Test plan

- [x] `pnpm --filter=@x402r/core exec vitest run tests/integration/partial-refund.fork.test.ts --project core:fork` — 4/4 pass (the existing 2 happy-path tests + the 2 new edge cases)
- [x] `pnpm check` (biome) — clean
- [x] `pnpm --filter=@x402r/core typecheck` — clean
- [x] Full `pnpm --filter=@x402r/core test:fork` shows no regressions vs origin/main baseline. The same 6-7 unrelated fork tests fail on both origin/main and this branch due to local anvil/prool RPC flakiness (`HttpRequestError` on `anvil_setCode`, not contract logic); the new tests are not among the flaky ones.

### DIFF
--- a/packages/core/tests/integration/partial-refund.fork.test.ts
+++ b/packages/core/tests/integration/partial-refund.fork.test.ts
@@ -319,3 +319,72 @@ describe('Edge case: double capture+void on same paymentInfo', () => {
     expect(payerBalanceAfter).toBe(payerBalanceBefore)
   }, 60_000)
 })
+
+// ---------------------------------------------------------------------------
+// Edge case (d): capture overspend — capture(amount > capturableAmount)
+// ---------------------------------------------------------------------------
+
+// Pins down on-chain behavior when a merchant tries to capture MORE than the
+// authorized capturableAmount (e.g., off-by-one in their amount math, or a
+// stale read of capturableAmount before another capture drained it). The SDK
+// path (`packages/core/src/actions/payment/capture.ts:17-34`) wraps
+// `writeContract` with no client-side amount validation, so the behavior is
+// purely contract-defined.
+describe('Edge case: capture overspend (amount > capturableAmount)', () => {
+  it('reverts and leaves capturableAmount unchanged', async () => {
+    // Fresh paymentInfo with salt: 5n to avoid colliding with Scenarios above.
+    const overspendInfo: PaymentInfo = { ...paymentInfo, salt: 5n }
+    const { collectorData, tokenCollector } = await createCollectorData(
+      anvilBaseSepolia.getWalletClient(testRoles.payer.address),
+      overspendInfo,
+    )
+    const authHash = await payerClient.payment.authorize(
+      overspendInfo,
+      DEFAULT_AMOUNT,
+      tokenCollector,
+      collectorData,
+    )
+    await publicClient.waitForTransactionReceipt({ hash: authHash })
+
+    await testClient.increaseTime({ seconds: ESCROW_FAST_FORWARD })
+    await testClient.mine({ blocks: 1 })
+
+    const amountsBefore = await merchant.payment.getAmounts(overspendInfo)
+    expect(amountsBefore.capturableAmount).toBe(DEFAULT_AMOUNT)
+
+    // Snapshot receiver USDC balance — if the overspend silently let an
+    // out-of-range amount through, receiver tokens would move.
+    const receiverBalanceBefore = await publicClient.readContract({
+      address: USDC,
+      abi: erc20Abi,
+      functionName: 'balanceOf',
+      args: [testRoles.receiver.address],
+    })
+
+    // Attempt to capture MORE than authorized. SDK returns a tx hash (no
+    // pre-simulation), on-chain reverts at receipt level (same pattern as
+    // edge cases (b) and (c)).
+    const overspendHash = await merchant.payment.capture(
+      overspendInfo,
+      DEFAULT_AMOUNT + 1n,
+      '0x',
+    )
+    const overspendReceipt = await publicClient.waitForTransactionReceipt({
+      hash: overspendHash,
+    })
+    expect(overspendReceipt.status).toBe('reverted')
+
+    // State invariants: capturableAmount unchanged, receiver tokens didn't
+    // move. Together these prove the contract gate held.
+    const amountsAfter = await merchant.payment.getAmounts(overspendInfo)
+    expect(amountsAfter.capturableAmount).toBe(DEFAULT_AMOUNT)
+
+    const receiverBalanceAfter = await publicClient.readContract({
+      address: USDC,
+      abi: erc20Abi,
+      functionName: 'balanceOf',
+      args: [testRoles.receiver.address],
+    })
+    expect(receiverBalanceAfter).toBe(receiverBalanceBefore)
+  }, 60_000)
+})
--- a/packages/core/tests/integration/partial-refund.fork.test.ts
+++ b/packages/core/tests/integration/partial-refund.fork.test.ts
@@ -160,3 +160,117 @@ describe('Scenario 3: Partial capture + void remainder', () => {
     expect(payerBalanceAfter - payerBalanceBeforeVoid).toBe(REFUND_AMOUNT)
   }, 60_000)
 })
+
+// ---------------------------------------------------------------------------
+// Edge case (b): voidPayment after full capture
+// ---------------------------------------------------------------------------
+
+// Pins down on-chain behavior when a merchant calls `voidPayment()` after
+// already capturing the full authorized amount (capturableAmount === 0). The
+// JSDoc on `voidPayment` (packages/core/src/actions/refund-budget/voidPayment.ts:17-18)
+// describes the call as "empties the authorization in one transaction" but
+// doesn't specify what happens when there's nothing left to empty. This test
+// documents the actual contract behavior so a future contract change can't
+// silently change the SDK's surface.
+describe('Edge case: voidPayment after full capture', () => {
+  it('reverts (or no-ops) when capturableAmount is already 0', async () => {
+    // Fresh paymentInfo with salt: 4n to avoid colliding with Scenario 3's
+    // salt: 3n (same payer + receiver + asset means salt is the only
+    // disambiguator in the escrow nonce).
+    const edgePaymentInfo: PaymentInfo = { ...paymentInfo, salt: 4n }
+    const { collectorData, tokenCollector } = await createCollectorData(
+      anvilBaseSepolia.getWalletClient(testRoles.payer.address),
+      edgePaymentInfo,
+    )
+    const authHash = await payerClient.payment.authorize(
+      edgePaymentInfo,
+      DEFAULT_AMOUNT,
+      tokenCollector,
+      collectorData,
+    )
+    await publicClient.waitForTransactionReceipt({ hash: authHash })
+
+    await testClient.increaseTime({ seconds: ESCROW_FAST_FORWARD })
+    await testClient.mine({ blocks: 1 })
+
+    // Capture the full DEFAULT_AMOUNT so capturableAmount drains to 0.
+    const captureHash = await merchant.payment.capture(
+      edgePaymentInfo,
+      DEFAULT_AMOUNT,
+      '0x',
+    )
+    await publicClient.waitForTransactionReceipt({ hash: captureHash })
+
+    const amounts = await merchant.payment.getAmounts(edgePaymentInfo)
+    expect(amounts.capturableAmount).toBe(0n)
+
+    // Observed behavior (pinned 2026-05-18 against the deployed
+    // PaymentOperator + EscrowPeriod fixtures): the SDK's `voidPayment()`
+    // submits the tx and the wallet client returns a tx hash without throwing
+    // (`writeContract` on anvil doesn't simulate before sending). On-chain,
+    // the `void()` call REVERTS — see the receipt status below. The
+    // capturableAmount stays at 0n (no state change), confirming nothing was
+    // double-voided.
+    //
+    // This pins down the contract-side guarantee: if a future contract
+    // version turns the revert into a no-op success, or starts simulating
+    // pre-send (so the SDK throws), this test will fail and force a
+    // deliberate decision.
+    const voidHash = await merchant.payment.voidPayment(edgePaymentInfo)
+    const voidReceipt = await publicClient.waitForTransactionReceipt({
+      hash: voidHash,
+    })
+    expect(voidReceipt.status).toBe('reverted')
+
+    const amountsAfter = await merchant.payment.getAmounts(edgePaymentInfo)
+    expect(amountsAfter.capturableAmount).toBe(0n)
+  }, 60_000)
+})
+
+// ---------------------------------------------------------------------------
+// Edge case (c): double-application / nonce reuse
+// ---------------------------------------------------------------------------
+
+// Depends on the file-level `beforeAll` paymentInfo (salt: 3n) being
+// capture+void'd by the preceding `Scenario 3: Partial capture + void
+// remainder` describe block. Vitest runs `describe` blocks in file order, so
+// by the time this runs, `paymentInfo` has been fully consumed (authorized,
+// partial-captured, then voided). Reusing the same paymentInfo here exercises
+// the replay-protection boundary: the escrow nonce (see
+// `computeEscrowNonce` at packages/core/src/payment/hashing.ts:87-97) is
+// derived from `(chainId, escrowAddress, paymentInfo with payer=0)`, so a
+// second authorize with the same salt should collide with the already-spent
+// nonce and revert.
+describe('Edge case: double capture+void on same paymentInfo', () => {
+  it('rejects second authorize on the same salt (nonce collision)', async () => {
+    const { collectorData, tokenCollector } = await createCollectorData(
+      anvilBaseSepolia.getWalletClient(testRoles.payer.address),
+      paymentInfo,
+    )
+
+    // Observed behavior (pinned 2026-05-18): the SDK's `authorize()` submits
+    // the tx and returns a hash without throwing (no pre-send simulation).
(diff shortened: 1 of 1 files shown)


vraspar:
### SDK Review

Found 2 issues (axes: tests, coverage):

1. **[Tests]** Cross-describe nonce dependency is doc-only — falsely greens under `--shuffle` or `.only`. Edge case (c) reuses the file-level `paymentInfo` (salt: 3n) and relies on Scenario 3 having already consumed it. The dependency lives in a comment but isn't asserted by the test. Under `vitest --shuffle` or `.only` isolation, the authorize would succeed (fresh nonce) and the receipt would be `'success'` — silently mis-pinning the security boundary. Fix: add `expect((await merchant.payment.getAmounts(paymentInfo)).capturableAmount).toBe(0n)` at the top of the `it` body, before the replay authorize.

   https://github.com/BackTrackCo/x402r-sdk/blob/eeb69ca878a8252bccd415ff26e626209c69b5f4/packages/core/tests/integration/partial-refund.fork.test.ts#L252-L261

2. **[Coverage]** Missing symmetric case: `capture(amount > capturableAmount)` overspend. The PR pins the void-after-full case (b) and the replay case (c). The symmetric "capture more than authorized" path has no fork test — `capture.ts` wraps `writeContract` with no SDK-side amount validation, so behavior is purely contract-defined and undocumented in test form. Sits on the same axis as edge case (b) ("called when nothing left to operate on") and fits the same describe block.

   https://github.com/BackTrackCo/x402r-sdk/blob/eeb69ca878a8252bccd415ff26e626209c69b5f4/packages/core/src/actions/payment/capture.ts#L17-L34

**On the three plan-doc cases** (`AUTHCAPTURE_SDK_MIGRATION.md:66`): `merchantAmount === 0` is deferred (reasonable — needs contracts-side investigation first), `merchantAmount === capturableAmount` is covered by edge case (b), and double-application is covered by edge case (c). The deferred case is the only gap from the plan; the `capture` overspend above is the one new gap surfaced by reading neighboring production code.

---

vraspar:
Thanks for both findings — addressed in 34468ea and ac43622.

**Finding 1 (precondition assertion):** Added the explicit `expect(preAmounts.capturableAmount).toBe(0n)` guard at the top of the nonce-reuse test (commit 34468ea). One framing clarification, though: under `--shuffle` or `.only` the test wouldn't have silently greened — the existing assertion

```ts
expect(replayReceipt.status).toBe('reverted')
```

would have failed loudly (a fresh `paymentInfo` would make the second authorize succeed with status `'success'`, not `'reverted'`). So this isn't a silent-pass fix. The value of the precondition is defense-in-depth: it makes the cross-describe dependency executable instead of comment-only, and converts a confusing "expected reverted, got success" downstream failure into a clear "expected capturableAmount 0n, got [full]" failure at the top of the test. Worth doing regardless — just wanted to note the risk model.

**Finding 2 (capture overspend):** Valid gap — `packages/core/src/actions/payment/capture.ts:17-34` has no SDK-side amount validation, so contract behavior on `amount > capturableAmount` was undocumented in test form. Added as Edge case (d) in commit ac43622 with snapshot + revert + state-invariant + receiver-balance-no-movement assertions, matching the pattern of (b) and (c).

Targeted run: all 5 tests in `partial-refund.fork.test.ts` pass.

A1igator:
### SDK Review — Round 2

Found 0 unresolved blockers since commit `eeb69ca` (axes: tests).

- Cross-describe nonce-reuse precondition assertion — now at `partial-refund.fork.test.ts:269-270` (commit `34468ea`).
- Symmetric `capture(amount > capturableAmount)` overspend test — now at `partial-refund.fork.test.ts:333-390` (commit `ac43622`).

**Still open:** none.

**New:** none above the blocker threshold.

**Informational (not blocking, not requesting changes):**

- The comment at `packages/core/tests/integration/partial-refund.fork.test.ts:330` cites `packages/core/src/actions/payment/capture.ts:17-34`. That path doesn't exist — `capture.ts` lives at `packages/core/src/actions/operator/capture.ts` (line range 17-34 is correct for that file). Same stale path is in the PR description. Worth a one-line fix on the next push, but not blocking.

   https://github.com/BackTrackCo/x402r-sdk/blob/ac43622bded989c5b51c8edffb8b823f08c80885/packages/core/tests/integration/partial-refund.fork.test.ts#L328-L332

---

vraspar:
Fixed the stale path citation in commit `bf68431` — the comment in `partial-refund.fork.test.ts:330` now correctly references `packages/core/src/actions/operator/capture.ts:17-34`. Left my earlier reply comment on this PR alone since editing posted comments creates more noise than it removes; the source-of-truth in the code is what matters going forward. Thanks for the catch.

A1igator:
### SDK Review — Round 3

Found 0 unresolved blockers since commit `ac43622` (axes: tests).

- Stale `capture.ts` source path in edge case (d) comment — fixed to `packages/core/src/actions/operator/capture.ts` (commit `bf68431`).

**Still open:** none.

**New:** none.

---

A1igator (dismissed):

vraspar:
Follow-up notes from a deeper read. Three findings + two trivial improvements + one suggested addition, plus backlog pointers.

### Findings

**1. The "writeContract doesn't simulate" pattern is viem default, not anvil-specific.** The inline comments (lines 209, 277, 364) frame this as anvil behavior, but `walletClient.writeContract` has the same property on Base mainnet — verified zero `simulateContract` calls across `packages/core/src/actions/{operator/authorize.ts,operator/capture.ts,refund-budget/voidPayment.ts}`. The implication is bigger than the comments suggest: a merchant integration that trusts `await payment.voidPayment(...)` to throw on failure will silently miss reverts in production. Queued for SDK JSDoc + scenarios README in PR 3 of the publish-readiness series.

**2. Case (c) names the wrong nonce layer.** Inline comment at `partial-refund.fork.test.ts:281-282` says "the escrow nonce (see computeEscrowNonce at hashing.ts:87-97) is already consumed". The actual replay-protection gate is **USDC's ERC-3009 `authorizationState` mapping**, not the escrow contract. `computeEscrowNonce` only produces the deterministic nonce value; the escrow doesn't maintain its own consumption map. Test is correct (replay does fail), but it fails at a different layer than the comment claims.

**3. Case (b)'s payer-balance assertion at line 244 is tautological in context.** Between full capture (moves escrow→receiver) and the reverting void, nothing can touch payer balance. Defensible as a future-regression guard, but not catching anything today. Worth a one-line code comment to that effect so future readers don't assume it's an active gate.

### Trivial improvements (recommend folding into this PR)

- Fix the case (c) inline comment: "escrow nonce" → "ERC-3009 nonce in USDC's `authorizationState`".
- Add `expect(receipt.logs).toHaveLength(0)` to all three new tests. Receipt-status + balance-unchanged can't catch the "event emitted but state reverted" partial-state bug. `PaymentVoided`/`PaymentAuthorized`/`PaymentCaptured` events at `packages/core/src/abis/generated.ts:686+`.

### Suggested addition (fold if scope allows)

- **Void during escrow window** — capture is time-gated via `EscrowPeriod.CAPTURE_PRE_ACTION_CONDITION`, void is not (`PaymentOperator.sol:323-327` shows `VOID_PRE_ACTION_CONDITION` defaults to `address(0)`). The asymmetry isn't tested or documented. ~30 LOC, fresh salt (6n), same fixture as cases (b)/(d). Highest-value missing test in this neighborhood.

### Backlog (post-0.3.0, filed in `x402r-notes/plans/AUTHCAPTURE_SDK_MIGRATION.md`)

- Replay-with-different-amount in case (c) — pins ERC-3009-layer vs amount-validation order
- Sequential overspend after partial capture — boundary at `capturableAmount + 1` exactly
- Void on never-authorized `paymentInfo` — different SDK path, same `ZeroAuthorization` revert
- USDC nonce direct check via `authorizationState(payer, nonce)` — stronger than balance-proxy

vraspar:
All four groups from your review are in — six commits on top of bf68431, all in `partial-refund.fork.test.ts`. File goes from 5 to 10 tests.

### Findings (1-3)

- **F1 (viem default, not anvil):** rewrote the inline comments at the three locations to attribute the no-pre-simulation to viem's `walletClient.writeContract` (verified zero `simulateContract` calls across `packages/core/src/actions/{operator,refund-budget}/`) and flagged the merchant-integration implication. Commit `4e1b807`.
- **F2 (case c wrong nonce layer):** corrected the comment to name USDC's ERC-3009 `authorizationState(payer, nonce)` as the actual gate, with `computeEscrowNonce` noted as the derivation. Same commit.
- **F3 (case b tautological assertion):** your framing got incorporated as a one-line comment above the assertion, not removed — kept as defense-in-depth for future test-ordering changes. Same commit.

### Trivial improvements

- `expect(receipt.logs).toHaveLength(0)` added to cases (b), (c), (d). Commit `011c19e`.

### Suggested addition

- **Edge case (e)** void during escrow window — succeeds (status `'success'`, not reverted), capturableAmount drains to 0, payer recovers full DEFAULT_AMOUNT (no capture, no fees). Commit `c0ec616`.

### Backlog (all four pulled forward)

- **(f) different-amount replay** in case (c) describe — pins the ERC-3009 nonce gate as amount-agnostic. Commit `bf57247`.
- **(i) USDC `authorizationState` direct check** — reads the bool from USDC at `0x036CbD53842c5426634e7929541eC2318f3dCF7e` (Base Sepolia), asserts `true` after Scenario 3 lifecycle. Stronger than the balance-unchanged proxy. Same commit as (f).
- **(g) sequential overspend after partial capture** — boundary at `capturableAmount + 1n` after state mutation, distinct from (d)'s fresh-state version. Commit `996bdf6`.
- **(h) void on never-authorized paymentInfo** — different SDK path than (b), no prior escrow state, same receipt-level revert shape. Commit `6d7b70b`.

Targeted suite green (10/10). Full fork suite shows the pre-existing `anvil_setCode 400` flakies on other files — not regressions.

A1igator:
### SDK Review — Round 4

Found 0 unresolved blockers since commit `bf68431` (axes: tests, conventions).

**Resolved since last round:** n/a (no open issues from Round 3).

**Still open:** none.

**New:** none above the blocker threshold.

**Informational (not blocking, not requesting changes):**

- Edge case (i) at `partial-refund.fork.test.ts:419-444` reads USDC's `authorizationState(payer, escrowNonce)` and asserts `consumed === true`. The test inherits Scenario 3's nonce-consumption state but lacks the `expect(preAmounts.capturableAmount).toBe(0n)` precondition that (c) at line 285-286 and (f) at line 363-364 both carry. Under `--shuffle` or `.only`, (i) would fail with "expected true received false" rather than the clearer precondition diagnostic the rest of the describe block uses. Consistency-only — not a correctness issue.

   https://github.com/BackTrackCo/x402r-sdk/blob/6d7b70b0afc0232231a384ecdb66fe6fff0a7f67/packages/core/tests/integration/partial-refund.fork.test.ts#L419-L444

**Ruled out** (verified <75): inline ABI in (i) is fine (USDC isn't in `generated.ts`); the zero-log assertion is tautological under EVM top-level-revert atomicity but harmless; (e) not testing the capture-side of the asymmetry is incompleteness, not a convention violation.

---

vraspar:
Fresh review pass — three small items, none blocking.

**Medium**

- Case (e) void-during-escrow-window (success path) doesn't assert `logs.length > 0` — the `PaymentVoided` event should fire on a successful void. Would catch a bug where the void succeeds at state level but doesn't emit. Mirror of the `logs.length === 0` pattern used on the reverting cases.

**Low**

- Case (i) inlines the `authorizationState` ABI instead of importing the standard ERC-3009 ABI. Defensible (one call, one usage), but importing from viem or a USDC ABI would catch any future ABI drift.

**Nit**

- Case (c) inline comment near line 300 says "voidPayment" but the context is `authorize` — small wording fix.

8-case coverage is solid; salt allocation is clean; the comment-tightening commits landed all the corrections cleanly.

vraspar:
Round 5 polish — four small commits, all in `packages/core/tests/integration/partial-refund.fork.test.ts`:

**`9b251d6`** — case (i) precondition consistency (A1igator, Round 4): added `expect(preAmounts.capturableAmount).toBe(0n)` at the top of the case (i) `it` body, matching the same guard already on case (c) and case (f). Under `--shuffle` or `.only`, the test now fails fast with an actionable diagnostic instead of the downstream "expected true received false" from the `authorizationState` check.

**`01af091`** — case (e) success-path event-emission (vraspar): added `expect(voidReceipt.logs.length).toBeGreaterThan(0)` after the `status === 'success'` assertion. Mirrors the `logs.length === 0` pattern on the reverting cases. Catches the silent-no-emit regression class (e.g., a hook swallowing the post-action event) without locking in a brittle topic match.

**`2bb4315`** — ERC-3009 ABI standardization (vraspar): hoisted the inline `authorizationState` ABI literal in case (i) to a module-scope `erc3009Abi` constant. Checked viem and the SDK — neither exposes an ERC-3009 ABI export, so the constant lives in the test file with a spec link. Only one usage today, but a single source of truth if future tests assert nonce-consumption at the token layer.

**`e162b74`** — case (c) wording (vraspar): fixed `await payment.voidPayment(...)` → `await payment.authorize(...)` in the case (c) describe block (around line 300). The surrounding context is the authorize-replay assertion, not the void path.

All 10 tests still pass on `--project core:fork`. Pre-commit hooks clean on each commit.

vraspar:
Round 7 follow-up: tightened case (e)'s success-path event assertion to filter for `PaymentVoided` specifically via `parseEventLogs`, addressing the asymmetry vs the revert-side `toHaveLength(0)` exact-count checks. The old `logs.length > 0` passed on the silent-drop bug class (a successful void emits a USDC `Transfer` independently of `PaymentVoided`).

One surprise: `PaymentVoided` lives on `authCaptureEscrowAbi` (line 686), not `paymentOperatorAbi` — the operator emits `VoidExecuted`. Filtering against the operator ABI returned 0 events and failed the assertion. Switched to the escrow ABI; all 10 tests pass.

Commit: `0eeed33`.

A1igator:
### SDK Review — Round 5

Found 0 unresolved blockers since commit `6d7b70b` (axes: tests, conventions).

- Round 4 informational note: case (i) precondition assertion now in place at `partial-refund.fork.test.ts:419-426` (commit `9b251d6`), matching the (c)/(f) pattern.

**Still open:** none.

**New:** none above the blocker threshold.

**Verified during this round:**
- `parseEventLogs({ abi: authCaptureEscrowAbi, eventName: 'PaymentVoided', logs })` call shape matches viem v2.47.1; `PaymentVoided` is on the escrow ABI at `generated.ts:686` (operator emits `VoidExecuted` at line 3565); EVM inner-call log propagation guarantees the escrow's `PaymentVoided` reaches the operator's tx receipt — assertion is sound.
- Typo fix in case (c) did not touch case (b)'s `voidPayment` or case (d)'s `capture` method names — both still correct.
- Hoisting `erc3009Abi` to module scope is a wash at one consumer but doesn't violate any convention.

---

A1igator (approved):

PR #129, opened 2026-05-19
ci(security): release pipeline phase 0 — security baseline

vraspar:
## Summary

Phase 0 of the release-pipeline rollout (full plan: `x402r-notes/plans/RELEASE_PIPELINE.md`). Establishes the security baseline before npm Trusted Publishing + CI release automation lands in subsequent PRs.

- **`.github/CODEOWNERS`**: gate release-infrastructure paths on maintainer review (`/.github/`, root + workspace `package.json`, `pnpm-lock.yaml`, `.changeset/config.json`, `.npmrc`, `.nvmrc`, `RELEASING.md`, `scripts/`, `turbo.json`). General code changes still use normal review.
- **`.github/dependabot.yml`**: weekly npm + github-actions updates, grouped by dep-type to keep PR volume manageable. Workspace-aware via `directories: ["/", "/packages/*"]`. No auto-merge — every PR is reviewed.
- **`.npmrc`**: explicit `enable-pre-post-scripts=false`. pnpm 10 default, declared in-repo so the security stance is visible and protected against accidental override.
- **`.nvmrc`**: pin Node 22 to match the `engines` declaration and existing CI matrix.

## Why now

Coming PRs add the actual release workflow (`release.yml`, `pkg-pr-new.yml`). Those need CODEOWNERS active first so a malicious PR can't ship publish-altering changes without explicit maintainer approval. This PR is intentionally tiny and reviewable in 2 minutes.

## Test plan

- [ ] CI passes on the branch (lint, build, typecheck, test, fork tests, publish-validation, abi-drift).
- [ ] Dependabot runs its first scan within ~24h after merge (no action needed on our side; it starts on schedule).
- [ ] Touching `/.github/`, manifests, or release-relevant files in a future PR triggers a CODEOWNERS review request to `@vraspar`.

## Plan reference

Full multi-phase release pipeline plan: https://github.com/BackTrackCo/x402r-notes/blob/main/plans/RELEASE_PIPELINE.md

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #130, opened 2026-05-19
ci(release): release.yml — Trusted Publishing via OIDC

vraspar:
## Summary

Phase 2 of the release-pipeline rollout (full plan: `x402r-notes/plans/RELEASE_PIPELINE.md`). Adds the stable-release workflow that publishes `@x402r/*` packages via npm Trusted Publishing on push-to-main, gated by the `npm-publish` GitHub environment.

**Replaces manual `npm publish` from a maintainer laptop.** No long-lived `NPM_TOKEN`.

## How it works

1. PRs add changesets (`.changeset/*.md`) describing version bumps + changelogs.
2. On merge to `main`, this workflow opens/updates a "Version Packages" PR that consumes pending changesets, bumps `package.json` versions, and regenerates `CHANGELOG.md`.
3. Merging that Version PR triggers a publish run.
4. The publish run pauses at the `npm-publish` environment for explicit maintainer approval.
5. On approval, npm authenticates via OIDC (no token), publishes packages with provenance attestations, and creates per-package GitHub releases (`@x402r/<pkg>@<version>` tags).

## Security posture

- Top-level `permissions: {}` default-deny; per-job permissions narrowed to `contents: write`, `pull-requests: write`, `id-token: write`.
- Every action SHA-pinned to immutable commits (mitigates tj-actions-style retroactive tag-move attacks). Dependabot will keep SHAs + trailing version comments current.
- `step-security/harden-runner` in audit mode initially — surfaces actual egress profile across the first runs; Phase 7 will flip to `block` with the observed allow-list.
- `pnpm install --ignore-scripts` blocks dependency lifecycle hooks (Shai-Hulud-style postinstall worms have nowhere to run).
- Explicit `npm install --global npm@latest` because Node 22 ships with npm 10.x and OIDC Trusted Publishing requires npm ≥ 11.5.1.

## Prerequisites (already complete)

- ✅ `npm-publish` GitHub environment created with required-reviewer rule (Phase 1).
- ✅ Trusted Publisher configured on each of `@x402r/{core,sdk,helpers,cli}` pointing at `release.yml` + `npm-publish` environment (Phase 1).
- ⏳ Phase 0 PR (#129) merged for CODEOWNERS to gate this file in future edits.

## Test plan

This workflow only runs on push-to-main, so this PR's CI only exercises the existing `ci.yml`. Real validation happens in Phase 4 (first CI-driven publish):

- [ ] CI checks pass on this PR (no changes to `ci.yml`).
- [ ] After merge, observe the workflow run shows the `npm-publish` environment gate.
- [ ] Phase 4 will exercise the full flow: consume changesets → Version PR → merge → environment approval → publish → verify provenance via `npm audit signatures`.

## Plan reference

Full multi-phase release pipeline plan: https://github.com/BackTrackCo/x402r-notes/blob/main/plans/RELEASE_PIPELINE.md

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,72 @@
+name: Release
+
+on:
+  push:
+    branches: [main]
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release:
+    name: Publish to npm
+    runs-on: ubuntu-latest
+    # Required reviewer gate — see Settings -> Environments -> npm-publish.
+    # npm Trusted Publishing rules are scoped to this environment + workflow filename.
+    environment: npm-publish
+    permissions:
+      contents: write       # Push tags + Version PR commits
+      pull-requests: write  # Open/update the "Version Packages" PR
+      id-token: write       # OIDC handshake for npm Trusted Publishing
+    steps:
+      # Egress monitoring. Starts in audit mode so the first runs surface
+      # exactly which endpoints we actually hit; Phase 7 flips this to
+      # `block` with the observed allow-list.
+      - name: Harden runner
+        uses: step-security/harden-runner@ab7a9404c0f3da075243ca237b5fac12c98deaa5  # v2.19.3
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          # Full history so changesets can generate complete changelogs.
+          fetch-depth: 0
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093  # v6.0.8
+
+      - name: Setup Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6.4.0
+        with:
+          node-version-file: .nvmrc
+          cache: pnpm
+          registry-url: https://registry.npmjs.org
+
+      # npm >= 11.5.1 is required for Trusted Publishing OIDC support.
+      # Node 22 ships with npm 10.x; upgrade explicitly.
+      - name: Upgrade npm
+        run: npm install --global npm@latest
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile --ignore-scripts
+
+      - name: Build packages
+        run: pnpm build
+
+      # When there are unconsumed changesets in .changeset/, opens/updates
+      # the "Version Packages" PR. When that PR is merged (consuming the
+      # changesets), the next run takes the `publish` branch and publishes
+      # each changed package via OIDC with automatic provenance, then
+      # creates per-package GitHub releases (@x402r/core@1.2.3 style tags).
+      - name: Create Release PR or publish to npm
+        uses: changesets/action@63a615b9cd06ba9a3e6d13796c7fbcb080a60a0b  # v1.8.0
+        with:
+          publish: pnpm exec changeset publish
+          version: pnpm exec changeset version
+          title: "chore: version packages"
+          commit: "chore: version packages"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,9 +46,12 @@ jobs:
           registry-url: https://registry.npmjs.org
 
       # npm >= 11.5.1 is required for Trusted Publishing OIDC support.
-      # Node 22 ships with npm 10.x; upgrade explicitly.
-      - name: Upgrade npm
-        run: npm install --global npm@latest
+      # Pin to a specific version (not @latest) — pulling latest on a
+      # privileged runner re-introduces the upstream-trust hole we close
+      # everywhere else. Dependabot or quarterly review bumps this.
+      # --ignore-scripts: don't run npm's own lifecycle scripts on install.
+      - name: Upgrade npm (Trusted Publishing requires >= 11.5.1)
+        run: npm install --global npm@11.14.1 --ignore-scripts
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --ignore-scripts
@@ -56,11 +59,23 @@ jobs:
       - name: Build packages
         run: pnpm build
 
+      # Upfront gate: publint + attw across all publishable packages.
+      # Without this, a per-package prepublishOnly failure can occur
+      # *after* earlier packages have already published, leaving a
+      # partial release.
+      - name: Validate package tarballs (publint + attw)
+        run: pnpm test:pkg
+
       # When there are unconsumed changesets in .changeset/, opens/updates
       # the "Version Packages" PR. When that PR is merged (consuming the
       # changesets), the next run takes the `publish` branch and publishes
-      # each changed package via OIDC with automatic provenance, then
-      # creates per-package GitHub releases (@x402r/core@1.2.3 style tags).
+      # each changed package via OIDC, then creates per-package GitHub
+      # releases (@x402r/core@1.2.3 style tags).
+      #
+      # Provenance: NPM_CONFIG_PROVENANCE=true plus the per-package
+      # publishConfig.provenance flag together guarantee Sigstore-backed
+      # attestations regardless of changesets/cli inference. Verify with
+      # `npm audit signatures @x402r/<pkg>` after a publish.
       - name: Create Release PR or publish to npm
         uses: changesets/action@63a615b9cd06ba9a3e6d13796c7fbcb080a60a0b  # v1.8.0
         with:
@@ -70,3 +85,4 @@ jobs:
           commit: "chore: version packages"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_CONFIG_PROVENANCE: "true"
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -4,6 +4,10 @@
   "description": "One-shot CLI for paying x402 / x402r endpoints. Wallet-agnostic: raw key, JSON-RPC signer, or custom module.",
   "type": "module",
   "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public",
+    "provenance": true
+  },
   "bin": {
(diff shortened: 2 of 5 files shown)


A1igator:
Reviewed alongside #129/#131 for npm-publish-readiness / supply-chain posture, with the x402 reference workflow (`coinbase/x402` → `publish_npm_*.yml`) as a comparison point.

**Verdict: structurally excellent, three real gaps to close before the first publish.**

### What's already right (lots of it)
- Top-level `permissions: {}` default-deny.
- Per-job least-privilege scoped to exactly what's needed (`contents: write`, `pull-requests: write`, `id-token: write`).
- Every action SHA-pinned with trailing version comment — Dependabot can update both atomically.
- `step-security/harden-runner` present (audit now, block in Phase 7).
- `pnpm install --frozen-lockfile --ignore-scripts` — kills postinstall-worm class of attacks (Shai-Hulud-style) before they can run on the privileged publish runner.
- OIDC Trusted Publishing — no long-lived `NPM_TOKEN` anywhere.
- `environment: npm-publish` with required-reviewer gate.
- `concurrency` group without `cancel-in-progress` — correct (never cancel a mid-publish).

### Gaps to close before flipping the switch

**1. `npm install --global npm@latest` is a supply-chain hole.**
Each run pulls whatever npm has just published into a privileged runner that's about to hold an OIDC token scoped to `@x402r/*`. A compromised npm release (or registry-side compromise) would auto-install on the next push to main. The x402 reference pins `npm@11.14.1` (`publish_npm_x402_core.yml` etc.) and lets Dependabot bump it — same defense as SHA-pinning actions.

```diff
-      - name: Upgrade npm
-        run: npm install --global npm@latest
+      - name: Upgrade npm (Trusted Publishing requires >= 11.5.1)
+        run: npm install --global npm@11.14.1 --ignore-scripts
```

**2. Provenance is claimed in the PR description but not actually wired up.**
The PR says "publishes packages with provenance attestations," but:
- No `NPM_CONFIG_PROVENANCE` env var in the workflow.
- No `publishConfig.provenance: true` in any of `@x402r/{core,sdk,helpers,cli}` package.json.
- `@changesets/cli` does *not* infer provenance from `id-token: write` alone — it requires either the env var or the per-package `publishConfig` flag.

So as written, you'd publish without attestations. Belt-and-suspenders fix:

```diff
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_CONFIG_PROVENANCE: "true"
```

And add to each publishable package.json:
```json
"publishConfig": { "access": "public", "provenance": true }
```

Then verify post-publish with `npm view @x402r/core --json | jq .dist.attestations` and `npm audit signatures`.

**3. No pre-publish sanity gate on the tarball.**
The repo already has `pnpm test:pkg` (= `publint --strict` + `attw`). Run it before `changesets/action` — catches malformed exports / type-resolution bugs *before* they hit npm, where the only remedy is `npm unpublish` within 72h on a not-yet-depended-on version.

```diff
       - name: Build packages
         run: pnpm build
+
+      - name: Validate package tarballs (publint + attw)
+        run: pnpm test:pkg
```

### Smaller items (not blocking)
- Consider running `pnpm test:unit` before publish. CI gates `main`, so risk is low, but a publish workflow that only builds is one branch-protection-bypass away from shipping a regression.
- `harden-runner` in audit mode means egress isn't actually blocked yet — anything compromised in the install chain could exfiltrate the short-lived OIDC token. Plan to flip to `block` in Phase 7 is the right move; consider doing it sooner if the audit-mode allow-list stabilizes quickly across the first 1–2 publishes.
- `@x402r/cli`'s `bin` entry (if any) should be exercised in `pnpm test:pkg` too — `publint` covers it.
- `changesets/action@v1.8.0` (SHA `63a615b9…`) is current; no advisories I'm aware of on that SHA.

### Suggested merge order
1. #131 (unblocks CI on main).
2. #129 as-is (optionally fix the `repo:` slug in `.changeset/config.json` first — surfaces in #130's first changelog).
3. #130 after the three gaps above are addressed — at that point this workflow matches or exceeds the x402 reference's posture.

vraspar:
Thanks @A1igator — careful review. Agreed on all three substantive points; pushed fixes in `cd473a6`.

### Addressed

**1. npm pin (no more `@latest`)**
```diff
- run: npm install --global npm@latest
+ run: npm install --global npm@11.14.1 --ignore-scripts
```
Pin matches the x402 reference. Also added `--ignore-scripts` on the global install per the same defense-in-depth principle. Dependabot bumps `github-actions` SHAs but doesn't touch this string — quarterly review will rotate.

**2. Provenance wired up (belt-and-suspenders)**
You're right that OIDC + `id-token: write` alone does not trigger provenance — npm CLI needs the explicit flag. Applied both:
- Workflow env: `NPM_CONFIG_PROVENANCE: "true"`
- Per-package `publishConfig: { access: "public", provenance: true }` in all four publishable packages

Either alone would work, but the combination guarantees attestations regardless of changesets/cli inference. Verifiable post-publish via `npm audit signatures @x402r/<pkg>`.

**3. Pre-publish tarball gate**
Added `pnpm test:pkg` step before `changesets/action`:
```diff
+ - name: Validate package tarballs (publint + attw)
+   run: pnpm test:pkg
```
Prevents the partial-publish hazard where pkg A's `prepublishOnly` passes, pkg B's fails, and A is already on npm.

### Skipping (with reasoning)

- **`pnpm test:unit` before publish** — the Version PR's CI on the exact commit gates this before merge, so adding it to `release.yml` is redundant. You marked it non-blocking; keeping the publish workflow tight.
- **Harden-runner audit → block sooner** — sticking with Phase 7 of the plan. Want to observe the actual egress profile across the first 1–2 publishes (which sigstore endpoints get hit, etc.) before locking down the allow-list. Should be quick to flip once observed.

### One residual item (genuine follow-up, not addressed here)
`pnpm test:pkg` today is `pnpm publint && pnpm attw`, and both of those scripts in root `package.json` only filter `core`, `helpers`, `sdk` — `@x402r/cli` is missing from the upfront gate. Its `prepublishOnly` still runs `publint --strict && attw --pack` during `npm publish`, so cli isn't unvalidated — but it does mean the upfront gate's partial-publish protection doesn't cover cli. Worth a small follow-up PR to add cli to the `publint`/`attw` workspace scripts; didn't want to scope-creep this one.

Also merged latest `main` to pick up #131.

A1igator:
Second-pass review on `cd473a6` — all three blockers from my earlier review are addressed (npm pinned to `11.14.1`, provenance wired belt-and-suspenders via env + `publishConfig`, `pnpm test:pkg` upfront gate). Side-by-side with the x402 reference (`coinbase/x402/.github/workflows/publish_npm_*.yml`), this workflow is **stricter** — x402 has no `harden-runner`, no default-deny top-level `permissions:`, no upfront publint/attw gate, and uses one workflow per package (broader attack surface) instead of a single orchestrated changesets workflow.

Ready to merge once #129 lands. Filing **non-blocking** observations for the follow-up backlog:

1. **`pnpm/action-setup` reads version from `packageManager`**, not from an explicit `version:` arg. x402 sets `version: 11.1.1` directly in the workflow. Yours reads `pnpm@10.23.0` from `package.json`, which is CODEOWNERS-gated (#129), so the trust is the same. Pinning explicitly in the workflow shaves one hop. Optional.

2. **`prepublishOnly` already runs `publint + attw` per package**, so `pnpm test:pkg` is technically redundant with what `pnpm publish` does anyway. **The upfront gate is still load-bearing** — if package A passes `prepublishOnly` and package B fails mid-loop, A is already on npm. Keep the gate as-is; just noting the relationship.

3. **`.nvmrc` is `22` (latest 22.x patch).** For maximally deterministic publish runners, pin a patch (e.g., `22.11.0`). x402 doesn't do this either; minor.

4. **Commit signing on the "Version Packages" PR** isn't wired up. Branch protection requiring signed commits (settings-side) would catch this; the `github-actions[bot]` commits would need GPG/Sigstore signing. Defer to Phase 7 or later.

5. **Phase-4 post-publish verification** — once a real package goes out, verify Sigstore provenance is actually attached:
   ```sh
   npm audit signatures @x402r/core
   npm view @x402r/core --json | jq '.dist.attestations'
   ```
   `NPM_CONFIG_PROVENANCE=true` only fires when `npm publish` detects a supported OIDC environment, which Trusted Publishing satisfies. The wiring is correct; verify rather than trust.

6. **`harden-runner` in audit mode** — flip to `block` sooner rather than later once the audit run stabilizes (typically after 2-3 runs surface the egress profile). Each additional audit-mode run is a window where a compromised dependency could exfiltrate the short-lived OIDC token.

None of these block the merge.

vraspar:
Thanks for the second-pass and the side-by-side against the x402 reference @A1igator. Backlog items captured in `x402r-notes/plans/RELEASE_PIPELINE.md`:

| Item | Where it lives now |
|---|---|
| Explicit `pnpm version:` in workflow | Decided against — `packageManager` is CODEOWNERS-gated, same trust hop |
| `prepublishOnly` ↔ `test:pkg` relationship | Informational; upfront gate stays (partial-publish protection) |
| `.nvmrc` patch pin (e.g. `22.11.0`) | Decided against — marginal determinism gain; x402 doesn't pin patch either |
| **Commit signing on Version PR** | Folded into Phase 7 — research feasibility; `changesets/action` push via runner may need GitHub App with attached signing key to satisfy `required_signatures: true` |
| **Phase-4 post-publish verification** (`npm audit signatures`, `jq .dist.attestations`) | Folded into Phase 4 verification section — "verify, don't trust" |
| **Harden-Runner audit → block sooner** | Folded into Phase 7 — flip trigger relaxed from "one clean week" to "after first 1–2 publishes surface egress profile" |

Also need a formal "Approve" from you here once #129 lands (currently `REVIEW_REQUIRED` per `require_code_owner_reviews`). Plan note added so the merge-gate isn't a surprise on future PRs.

A1igator (approved):
All three Round-1 blockers addressed (npm pinned to 11.14.1 + `--ignore-scripts` on the global upgrade, provenance wired belt-and-suspenders via env + per-package `publishConfig`, upfront `pnpm test:pkg` partial-publish gate). Round-2 non-blockers triaged into Phase 4/7 of the plan or declined with reasoning.

Acknowledging the residual you self-flagged: root `pnpm publint`/`pnpm attw` skip `@x402r/cli`, so cli falls outside the upfront partial-publish protection — its own `prepublishOnly` still gates it during `npm publish`, so it's not unvalidated, just unprotected against a partial-publish state. Agreed that's a separate small follow-up PR rather than scope-creep here.

LGTM. Merge after #129 lands.

PR #143, opened 2026-05-19
ci: skip Codecov upload on Dependabot PRs

vraspar:
## Summary

- All 9 currently open Dependabot PRs are blocked on the `Test` required check.
- Root cause: codecov-action fails with `Token required because branch is protected` when `CODECOV_TOKEN` is empty, and Dependabot PRs run without secret access.
- Because `fail_ci_if_error: true`, the failed upload fails the whole `Test` job — even though the tests themselves pass.

## Fix

Gate the three coverage upload steps on `github.actor != 'dependabot[bot]'`. Dependabot PRs skip the upload entirely; coverage isn't useful on dep bumps anyway. Strict `fail_ci_if_error` is preserved for main pushes and regular contributor PRs, so a real Codecov outage still fails CI.

## Test plan

- [ ] CI on this PR passes `Test` (this PR is not from Dependabot, so the upload still runs and exercises the happy path).
- [ ] After merge, re-trigger CI on one of the open Dependabot PRs (e.g. #134) and verify `Test` is no longer failing on the upload step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,10 @@ jobs:
       - run: pnpm build
       - run: pnpm --filter=@x402r/core test:cov
       - name: Upload core coverage
+        # Dependabot PRs run without secret access, so CODECOV_TOKEN is empty
+        # and the upload fails with "Token required because branch is protected".
+        # Skip on dependabot — coverage uploads aren't useful on dep bumps anyway.
+        if: github.actor != 'dependabot[bot]'
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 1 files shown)


A1igator (approved):
Textbook fix — matches codecov-action's own recommended Dependabot carve-out, correctly avoids `pull_request_target` (which would hand secrets to untrusted dep code) and the Dependabot-secret alternative (same risk surface). `fail_ci_if_error: true` is preserved for `main` pushes and regular contributor PRs, so real Codecov outages still fail CI. LGTM.

### Follow-up (not in scope here)

`ci.yml` uses tag-pinned actions throughout (`actions/checkout@v4`, `pnpm/action-setup@v4`, `codecov/codecov-action@v5`, etc.). The supply-chain posture adopted in #130 — every third-party action SHA-pinned with trailing version comment — should eventually be brought to `ci.yml` too. A retroactive tag-move on any of those actions would land on every PR's CI without warning. Worth its own PR with Dependabot configured to keep the SHA+comment pair in sync (Phase 0's dependabot.yml already covers `github-actions` ecosystem, so the bumps will just work).

vraspar:
Replacing with proper OIDC-based fix in a follow-up PR. This patch only handled Dependabot; OIDC handles Dependabot + external fork contributors uniformly and eliminates the need for the long-lived CODECOV_TOKEN secret entirely. Matches the pattern used by microsoft/TypeScript.

PR #166, opened 2026-05-21
ci(release): rewrite release.yml as workflow_dispatch (kills auto-trigger noise)

vraspar:
## Why

Current `release.yml` triggers on every push to `main`, then pauses at a required-reviewer `npm-publish` environment gate. Result: every Dependabot bump, every unrelated commit creates a stale "Release" run sitting in the queue. Noise + no value.

**Research finding** (subagent on workflows from astro / vite / vuejs/core / changesets/changesets / turborepo / shadcn-ui / radix-ui / chakra-ui / trpc / react-router): **zero** popular projects use the "push + required-reviewer gate that pauses" pattern. Either no gate, or env gate only on the actual publish job. The pause-on-every-push was over-engineering on my part.

## What this PR does

**Two-job split** mirrors `changesets/changesets`'s `publish.yml` (their `version` + `publish` jobs gated by `hasChangesets == 'false'`). **Trigger** is `workflow_dispatch` rather than their `push: branches: [main, next]` — a deliberate deviation to eliminate the stale-auto-run problem from Dependabot pushes. Trade-off: an extra manual click per release in exchange for zero unsolicited workflow runs.

- **Trigger**: `workflow_dispatch:` only. Click "Run workflow" in the Actions UI when you want to release. Nothing else fires it.
- **`version` job**: no env, no OIDC, no publish secrets. Opens or updates the "Version Packages" PR. Outputs `hasChangesets` so the next job knows what to do.
- **`publish` job**: gated by `if: needs.version.outputs.hasChangesets == 'false'` — only runs after the Version PR was merged (on a previous dispatch). Has OIDC scope + provenance env.
- **One dispatch click does the right thing**: if changesets are pending, the version job updates the PR and publish skips. If no changesets pending, version exits as no-op and publish runs.

## Single-click experience (settings change you do once)

The `environment: npm-publish` label stays so npm Trusted Publishing's OIDC config still matches. **But** the required-reviewer protection rule on that environment can be removed — `workflow_dispatch` already gates on maintainer access; the env gate would just be a second confirmation click.

To go full single-click:
- Settings → Environments → `npm-publish` → uncheck Required reviewers → Save

Leave it on if you prefer belt-and-suspenders (two clicks per publish).

## What stays the same (security)

- All SHA-pinned actions (checkout v6.0.2, setup-node v6.4.0, pnpm-action-setup v6.0.8, harden-runner v2.19.3, changesets/action v1.8.0). Dependabot keeps SHA + version comment in sync.
- Top-level `permissions: {}` default-deny; per-job least-privilege.
- step-security/harden-runner in audit mode (Phase 7 flips to block once observed).
- `pnpm install --frozen-lockfile --ignore-scripts` blocks postinstall-worm class.
- `npm@11.14.1` pinned (no `@latest` on privileged runner).
- `persist-credentials: false` on every checkout — `GITHUB_TOKEN` doesn't leak into `.git/config` for downstream steps.
- OIDC Trusted Publishing (no `NPM_TOKEN`) + per-package `publishConfig.provenance: true` + `NPM_CONFIG_PROVENANCE: "true"` env. Belt-and-suspenders provenance.
- `pnpm test:pkg` upfront publint+attw gate before publish (prevents partial-publish hazard).
- `concurrency` without `cancel-in-progress` (never cancel mid-publish — leaves npm half-state).
- `timeout-minutes: 20` on every job.
- No `pull_request` trigger — fork PRs can't inject anything into this workflow, period.

## Reference

- Subagent research on 10 popular npm monorepo workflows (Option B from that triage).
- Plan: `x402r-notes/plans/RELEASE_PIPELINE.md`
- Operator runbook: `x402r-notes/sdk/RELEASING.md`

## Test plan

This workflow only runs on `workflow_dispatch`, so this PR's CI exercises just `ci.yml`. Real validation happens by **dispatching it manually after merge**:

- [ ] Merge this PR.
- [ ] Verify no auto-Release run fires on subsequent pushes (Dependabot PRs, your commits, etc.).
- [ ] Settings → Environments → `npm-publish` → decide whether to keep required-reviewer or drop it.
- [ ] When ready for first publish: do the prep work (workspace version reset + changeset reclassify + `pnpm changeset pre enter alpha`) in a separate PR, then click Run workflow → version → review the Version PR → merge it → click Run workflow → publish.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,29 +1,39 @@
 name: Release
 
+# Manually triggered only. Click "Run workflow" in the Actions UI when
+# you want to either open/update the Version Packages PR (if there are
+# pending changesets) or publish to npm (after the Version PR is merged).
+#
+# Changesets self-selects which mode to run via its `hasChangesets`
+# output — one click does the right thing.
+#
+# Industry pattern: changesets/changesets uses the same two-job split.
+# See x402r-notes/sdk/RELEASING.md for the operator runbook.
 on:
-  push:
-    branches: [main]
+  workflow_dispatch:
 
 permissions: {}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
+  # Never cancel a mid-release run — a partial publish leaves npm in an
+  # inconsistent state. Queue subsequent dispatches instead.
+  cancel-in-progress: false
 
 jobs:
-  release:
-    name: Publish to npm
+  # Job 1: open/update the "Version Packages" PR if there are pending
+  # changesets. Doesn't publish anything. No publish secrets are needed
+  # so no OIDC scope and no env gate.
+  version:
+    name: Version
     runs-on: ubuntu-latest
-    # Required reviewer gate — see Settings -> Environments -> npm-publish.
-    # npm Trusted Publishing rules are scoped to this environment + workflow filename.
-    environment: npm-publish
+    timeout-minutes: 20
+    outputs:
+      hasChangesets: ${{ steps.changesets.outputs.hasChangesets }}
     permissions:
-      contents: write       # Push tags + Version PR commits
-      pull-requests: write  # Open/update the "Version Packages" PR
-      id-token: write       # OIDC handshake for npm Trusted Publishing
+      contents: write       # Push the version-bump commit
+      pull-requests: write  # Open/update the Version Packages PR
     steps:
-      # Egress monitoring. Starts in audit mode so the first runs surface
-      # exactly which endpoints we actually hit; Phase 7 flips this to
-      # `block` with the observed allow-list.
       - name: Harden runner
         uses: step-security/harden-runner@ab7a9404c0f3da075243ca237b5fac12c98deaa5  # v2.19.3
         with:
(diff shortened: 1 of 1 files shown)


A1igator:
### Workflow Review (release.yml)

Workflow-focused review (outside the normal review-sdk scope; reviewed at user request). The two-job split, SHA pinning, `persist-credentials: false`, OIDC + environment scoping, default-deny `permissions: {}`, harden-runner (audit), `cancel-in-progress: false`, and the upfront `pnpm test:pkg` gate are all sound. Two observations:

1. **Publish job lacks a `refs/heads/main` ref guard.** `workflow_dispatch` lets the operator pick any ref from the branch dropdown. The job comment notes that trust is "scoped to this environment + workflow filename" — i.e., the trust does not bind to a ref. A dispatch from a stale or feature branch with the same workflow file and `npm-publish` environment would mint a valid OIDC token and publish from that ref. The established pattern across x402-ecosystem publish workflows is `if: github.ref == 'refs/heads/main'` on the publish job (22 of 23 upstream per-package publish workflows carry this guard). Adding the same condition to `version` would also reduce the risk surface on its `pull-requests: write` permission.

   https://github.com/BackTrackCo/x402r-sdk/blob/cb024522941fa1d9ddde1da91756b67be655a6c0/.github/workflows/release.yml#L80-L92

2. **`Pin npm` step in the `version` job has no purpose.** The version job only runs `changeset version` (file mutation) and the changesets action's PR-creation path. It never invokes `npm publish`, so the Trusted-Publishing-driven `npm >= 11.5.1` pin is unused. Removing the step shortens the job and avoids a global install on a runner that already holds `contents: write` + `pull-requests: write`.

   https://github.com/BackTrackCo/x402r-sdk/blob/cb024522941fa1d9ddde1da91756b67be655a6c0/.github/workflows/release.yml#L58-L62

---

vraspar:
Both review items addressed in \`0ceb027\`:

1. **Ref guard added to both jobs** — \`if: github.ref == 'refs/heads/main'\` on \`version\` and \`publish\`. Comment block explains the threat (workflow_dispatch ref dropdown isn't a security boundary) and references the 22/23 upstream pattern. The composition with \`needs.version\` is safe: skipped version yields empty-string \`hasChangesets\` output, which fails the \`== 'false'\` check, so publish skips on either guard.

2. **\`Pin npm\` removed from version job** — version only runs \`changeset version\` (file mutation) + opens PR via the action; no \`npm publish\` shell-out. Pin stays in publish job where it's needed for Trusted Publishing's npm >= 11.5.1 requirement. Verified against changesets/changesets' own \`publish.yml\` (their version job has no npm pin either).

Per a new repo-internal workflow rule, ran a review-chain subagent on this fold-in before push. It validated both fixes against actual upstream patterns (coinbase publish_npm_* family + changesets/changesets) and verified the \`needs:\` interaction. Verdict: APPROVE WITH MINOR (cosmetic comment-wording suggestions only; no blockers).

Re-approval needed before merge — \`dismiss_stale_reviews: true\` will have dropped the prior one on the new commit.

A1igator:
### SDK Review — Round 2

Both findings from round 1 are resolved (commit 0ceb027 since cb02452). No new issues.

**Resolved:**
- Publish job lacks a `refs/heads/main` ref guard → both `version` and `publish` now carry the guard.
- `Pin npm` step in the `version` job has no purpose → step removed with explanatory comment.

The composition note in the commit message checks out: skipped `needs.version` yields empty-string output, which fails the `== 'false'` check, so `publish` is gated by either guard. The redundant ref guard on `publish` is intentional defense-in-depth against future restructuring (e.g., later `if: always()` or dropped `needs:` line).

---

A1igator (approved):

PR #168, opened 2026-05-21
ci(release): drop cache: pnpm to close cache-poisoning path

vraspar:
## Summary

Drops `cache: pnpm` from both setup-node steps in `.github/workflows/release.yml` to eliminate a cross-workflow cache-poisoning vector.

## Threat model

- `ci.yml` runs on PRs (low privilege, no secrets) and uses `cache: pnpm`. It writes to the shared GitHub Actions pnpm cache from untrusted PR inputs.
- `release.yml` runs on workflow_dispatch from main (high privilege: `contents: write`, `id-token: write` for npm Trusted Publishing OIDC) and previously also used `cache: pnpm` — it would deserialize the same cache that PR runs populated.

Per Adnan Khan's GitHub Actions cache-poisoning research (https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/), a low-privilege workflow can poison a cache that a privileged workflow later reads, giving an attacker a path to inject malicious bytes into a release artifact.

## Upstream pattern (related, not identical)

`changesets/changesets`'s `publish.yml` uses `skip-cache: true` on their `ci-setup` composite in both jobs, with the comment `# avoid cache poisoning attacks`. Their threat is internal lateral movement between their own privileged workflows; ours is external PR-run cache poisoning the privileged release runner. **Same defensive pattern, related class of attack.** Since our workflow calls `actions/setup-node` directly rather than through a composite, the equivalent is omitting `cache: pnpm`.

Each setup-node step now carries an explanatory comment so the omission isn't mistaken for an oversight.

## Tradeoff

~10s of cold pnpm install per release run. Release is workflow_dispatch-only and infrequent, so the safety wins.

## Test plan

- [ ] Lint/format passes (biome check) — covered by pre-commit hook
- [ ] Next manual release.yml dispatch completes (cold install path)
- [ ] No structural changes to setup-node — `node-version-file: .nvmrc` and `registry-url` are preserved

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,11 +56,15 @@ jobs:
       - name: Setup pnpm
         uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093  # v6.0.8
 
+      # No `cache: pnpm` here: the privileged release runner must not
+      # deserialize the same cache that ci.yml writes to from
+      # untrusted PR runs (cache-poisoning vector). Upstream
+      # changesets/changesets uses `skip-cache: true` for the same
+      # reason. Cold install costs ~10s per release run.
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6.4.0
         with:
           node-version-file: .nvmrc
-          cache: pnpm
 
       # Note: no `Pin npm` step here. The version job only mutates files
       # via `changeset version` + opens a PR via the action — it never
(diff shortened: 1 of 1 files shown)


A1igator (approved):
### SDK Review

CI/security PR, no issues at threshold. Axes covered: workflow security audit, dead code, convention check, cross-reference against upstream.

**Verified**

- Threat-model premise holds: `ci.yml` runs on `pull_request:` with `cache: pnpm` across 7 jobs, so the shared-cache vector is real.
- Upstream citation accurate: `changesets/changesets` `publish.yml` uses `skip-cache: true # avoid cache poisoning attacks` on both `version` and `publish` jobs. The PR's comment mirrors the rationale.
- Independent corroboration: upstream x402 follows the same asymmetry. All 30+ npm publish workflows omit `cache:`, while CI workflows (`unit_tests.yml`, `check_*.yml`) keep `cache: pnpm`. Same shape as this PR.
- Mechanical: both `setup-node` steps have `cache: pnpm` removed; `node-version-file: .nvmrc` and `registry-url: https://registry.npmjs.org` (publish job) preserved. Comment indentation aligns with the step. Duplicated comment at both sites is intentional per the description.
- Full workflow audit: top-level `permissions: {}`, per-job least-privilege, all third-party actions pinned by SHA, `persist-credentials: false`, `--frozen-lockfile --ignore-scripts` on every install, `id-token: write` only on the publish job, `if: github.ref == 'refs/heads/main'` ref guard on both jobs, `cancel-in-progress: false`, `environment: npm-publish` for OIDC scoping, `pnpm test:pkg` (publint + attw) gate before publish, `NPM_CONFIG_PROVENANCE: "true"`.

**Non-blocking follow-up (separate change)**

`step-security/harden-runner` is in `egress-policy: audit` on the publish job. For the `id-token: write` runner that mints npm publish tokens, switching to `egress-policy: block` (after one audit run to derive the allowlist) would prevent OIDC-token exfiltration if a malicious package slipped past `--ignore-scripts` and `pnpm test:pkg`.

---

PR #185, opened 2026-06-02
fix(helpers): gate forwardToArbiter on AUTH_CAPTURE_SCHEME; align to renamed auth-capture graph

vraspar:
## What

`@x402r/helpers`' `forwardToArbiter` hook gated on the pre-rename `'authCapture'` scheme literal. The scheme was renamed `authCapture` → `auth-capture`, so once helpers consume the renamed `@x402r/evm`, the gate `requirements.scheme !== 'authCapture'` matched nothing and the merchant→arbiter forwarding would **silently early-return on every settlement**. CI didn't catch it because the test fixtures hardcoded the same old literal.

This gates on the imported `AUTH_CAPTURE_SCHEME` constant instead, so the comparison tracks the package's scheme constant rather than a hardcoded string.

## Why the dep bump is in the same PR

Importing `AUTH_CAPTURE_SCHEME` is only possible against `@x402r/evm@0.2.0-alpha.1` — the renamed build exposes the constant; `0.2.0-alpha.0` does not and still emits `'authCapture'`. And bumping to the renamed scheme *without* the literal fix is exactly what breaks the bridge. So the fix and the bump are one atomic change. The renamed `@x402r/evm` peer-requires the 2.14 graph, hence the peer/devDep bumps.

## Changes

**`@x402r/helpers`**
- `forwardToArbiter`: gate on `AUTH_CAPTURE_SCHEME` (imported from `@x402r/evm`) instead of `'authCapture'`.
- peerDeps: `@x402/core` `>=2.5.0` → `^2.14.0`; add `@x402/evm ^2.14.0` + `viem ^2.48.11`; `@x402r/evm` floor → `>=0.2.0-alpha.1`. devDeps bumped to match.
- JSDoc + README: `AuthCaptureServerScheme` → `AuthCaptureEvmScheme` (the real class), `authCapture` → `auth-capture` prose.
- Tests: fixtures migrated to the imported constant; **3 new regression cases** — fires for `AUTH_CAPTURE_SCHEME`, skips a different scheme, and skips the legacy `'authCapture'` literal (guards against a reintroduced hardcode).

**examples**
- Repair the auth-capture example: kebab `@x402r/evm/auth-capture/*` subpaths, upstream client from `@x402/evm/auth-capture/client`, `scheme: AUTH_CAPTURE_SCHEME`; bump example deps to the 2.14 graph.

## Validation

- `@x402r/helpers`: 30/30 tests pass (incl. the 3 regression cases), typecheck clean.
- Full-workspace `turbo build` + `typecheck` green; biome clean.
- The regression suite was verified to genuinely fail (7/10 cases) when the gate is reverted to the old hardcoded literal — it is not tautological.

## Note on examples typecheck

The `examples` package has pre-existing `tsc` errors (a viem 2.47↔2.50 dual-version split + unrelated SDK-API drift in files this PR does not touch). Examples run via `tsx` and are not in the typecheck CI gate. This PR *reduces* `http-wire-capture.ts` from 5 → 2 errors; the 2 residual are the pre-existing viem split. Fully reconciling that split means bumping the `@x402r/core`/`@x402r/sdk` viem pins — a separate step in the broader 2.14 alignment.

First step of aligning the SDK to the upstreamed auth-capture scheme.

### DIFF
--- a/.changeset/helpers-forwardtoarbiter-scheme-rename.md
+++ b/.changeset/helpers-forwardtoarbiter-scheme-rename.md
@@ -1,0 +1,15 @@
+---
+"@x402r/helpers": minor
+---
+
+Gate `forwardToArbiter` on the `AUTH_CAPTURE_SCHEME` constant from `@x402r/evm` instead of a hardcoded `'authCapture'` literal, and align `@x402r/helpers` to the renamed auth-capture scheme graph.
+
+**Why:** the scheme literal was renamed `authCapture` → `auth-capture`. The hardcoded gate matched nothing once helpers consumed the renamed scheme, so the merchant→arbiter forwarding would silently early-return on every settlement.
+
+**Breaking (peer requirements):**
+
+- `@x402/core` peer raised `>=2.5.0` → `^2.14.0`.
+- `@x402/evm ^2.14.0` and `viem ^2.48.11` added as peers.
+- `@x402r/evm` peer floor raised to `>=0.2.0-alpha.1` — the renamed build that exposes `AUTH_CAPTURE_SCHEME = "auth-capture"` and drops the local client subpath (the client is now consumed from `@x402/evm/auth-capture/client`).
+
+`forwardToArbiter` now fires only when `requirements.scheme === AUTH_CAPTURE_SCHEME` (`"auth-capture"`); it no longer matches the legacy `"authCapture"` literal.
--- a/examples/package.json
+++ b/examples/package.json
@@ -30,11 +30,11 @@
   "devDependencies": {
     "@types/express": "^5.0.0",
     "@types/node": "^25.9.0",
-    "@x402/core": "2.12.0",
-    "@x402/evm": "2.12.0",
-    "@x402/express": "2.12.0",
-    "@x402/fetch": "2.12.0",
-    "@x402r/evm": "0.2.0-alpha.0",
+    "@x402/core": "2.14.0",
+    "@x402/evm": "2.14.0",
+    "@x402/express": "2.14.0",
+    "@x402/fetch": "2.14.0",
+    "@x402r/evm": "0.2.0-alpha.1",
     "express": "^5.0.0"
   }
 }
--- a/examples/scenarios/README.md
+++ b/examples/scenarios/README.md
@@ -39,7 +39,7 @@ Demonstrates the atomic settlement path. In production the merchant advertises t
 
 2-role flow: authorize → capture(partial) → voidPayment().
 
-The new authCapture partial-refund pattern. Replaces the old single-tx `refundInEscrow(amount)` with a two-tx flow: merchant captures the amount they keep, then `voidPayment()` returns the remainder to the payer. No allowance setup, no ReceiverRefundCollector — the escrow handles it. Asserts payer net loss equals merchant-keep, receiver delta + fee delta equals merchant-keep.
+The new auth-capture partial-refund pattern. Replaces the old single-tx `refundInEscrow(amount)` with a two-tx flow: merchant captures the amount they keep, then `voidPayment()` returns the remainder to the payer. No allowance setup, no ReceiverRefundCollector — the escrow handles it. Asserts payer net loss equals merchant-keep, receiver delta + fee delta equals merchant-keep.
 
 ### permit2-charge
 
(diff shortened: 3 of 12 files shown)


A1igator:
### SDK Review

Found 0 blocking issues (axes: tests, conventions, dead code, example alignment, SDK design). The scheme-rename surface was verified against the published `@x402r/evm@0.2.0-alpha.1` and `@x402/evm@2.14.0` packages: `AUTH_CAPTURE_SCHEME` resolves to `"auth-capture"`, `EIP3009_TOKEN_COLLECTOR_ADDRESS` matches the pinned literal, the `auth-capture/{server,facilitator}` subpaths export `AuthCaptureEvmScheme`, the client now correctly resolves from upstream `@x402/evm/auth-capture/client`, and all `index.ts` re-exports exist on the alpha.1 root. Tests restore `globalThis.fetch` per case and the rename regression guard pins the value correctly.

Two sub-threshold notes (non-blocking):

1. **Comment redundancy** — the comment names the constant twice after the path→name edit: `Source: @x402r/evm EIP3009_TOKEN_COLLECTOR_ADDRESS (EIP3009_TOKEN_COLLECTOR_ADDRESS)`. Drop the trailing parenthetical.

   https://github.com/BackTrackCo/x402r-sdk/blob/25606dad5b780982398386b057db42b3d00316ce/examples/scenarios/http-wire-capture.ts#L136-L138

2. **Peers not imported directly** — `viem` and `@x402/evm` are added as hard `@x402r/helpers` peers, but `src/` imports neither directly (only `@x402r/evm`, `@x402r/core`, `@x402/core/server`). They are transitively required by `@x402r/evm`, so surfacing them is defensible; flagging only in case the intent was a direct dependency.

   https://github.com/BackTrackCo/x402r-sdk/blob/25606dad5b780982398386b057db42b3d00316ce/packages/helpers/package.json#L37-L41

---

A1igator (dismissed):

vraspar:
### SDK Review — Round 2

1 new issue + 1 nit since the round-1 review (commit 25606dad); code is unchanged since then.

Verified the import surface against the published packages: `@x402/evm@2.14.0` exports `./auth-capture/client`, and `@x402r/evm@0.2.0-alpha.1` root exports `AUTH_CAPTURE_SCHEME` (`"auth-capture"`) + `EIP3009_TOKEN_COLLECTOR_ADDRESS`. The PR's imports and the "now a root export" comment resolve correctly — no module-resolution break.

**Still open from round 1:** the doubled `EIP3009_TOKEN_COLLECTOR_ADDRESS (EIP3009_TOKEN_COLLECTOR_ADDRESS)` parenthetical; `viem` / `@x402/evm` peers not imported directly.

**New:**

1. **Example alignment** — `examples/scenarios/README.md:54` still documents the removed camelCase subpaths `@x402r/evm/authCapture/{server,facilitator,client}`. `http-wire-capture.ts` was migrated to kebab `/auth-capture/`, but the README describing that same scenario was not. The client path is doubly stale — it moved upstream to `@x402/evm/auth-capture/client`.

   https://github.com/BackTrackCo/x402r-sdk/blob/25606dad5b780982398386b057db42b3d00316ce/examples/scenarios/README.md#L54

2. **Tests (nit)** — `forward-to-arbiter.test.ts:178` `expect(AUTH_CAPTURE_SCHEME).toBe('auth-capture')` asserts the imported constant against the literal — a value canary, not a gate check. The behavioral assertion below it (spy not called for the legacy `'authCapture'`) is what guards the rename. Optional.

   https://github.com/BackTrackCo/x402r-sdk/blob/25606dad5b780982398386b057db42b3d00316ce/packages/helpers/tests/forward-to-arbiter.test.ts#L178

---

vraspar:
Rebased onto latest `main` and addressed the review notes. Summary of what changed and where I'm pushing back:

### Accepted / fixed
- **`examples/scenarios/README.md`** — corrected the stale camelCase subpaths: server/facilitator now `@x402r/evm/auth-capture/*`, and the client now correctly points at upstream `@x402/evm/auth-capture/client` (it moved upstream in 2.14). Also fixed the `authCapture` → `auth-capture` prose. It now matches `http-wire-capture.ts`'s actual imports — good catch, it was shipping copy-paste-broken paths.
- **Doubled comment** — removed the redundant `EIP3009_TOKEN_COLLECTOR_ADDRESS (…)` parenthetical in `http-wire-capture.ts`.

### Pushing back (kept as-is)
- **`expect(AUTH_CAPTURE_SCHEME).toBe('auth-capture')`** in `forward-to-arbiter.test.ts` — keeping it. It's not a gate check (correct), but it's the *only* assertion in the suite that pins the cross-package wire literal to its exact value. Every behavioral case is expressed in terms of the imported constant, so if a future `@x402/evm`/`@x402r/evm` bump silently changed the constant's value, they'd all still pass green — this one-line canary is what catches that drift.
- **`viem` / `@x402/evm` as helpers peers** — leaving them for now. They're transitively required via `@x402r/evm`'s own peers, so surfacing them is defensible; dropping them and letting `@x402r/evm` govern is also fine. I'd rather let the (now re-running) `knip --strict` check tell us whether it considers them unused before trimming, instead of guessing.

### CI note
The red "Publish Validation" was a `harden-runner` provisioning stall (~278s in pre-run vs ~5s on every sibling job in the same run) that ate the job's 5-minute timeout before `pnpm build` finished — the job was *cancelled*, not a real validation failure (it passes on `main`, and this PR touches no workflows or publish logic). The rebase push re-runs it fresh. Bumping that job's tight timeout is a separate infra nit.

### Out of scope (flagging)
`packages/cli/src/pay/index.ts` still imports `registerAuthCaptureEvmScheme` from the dropped camelCase `@x402r/evm/authCapture/client` subpath — pre-existing on `main`, and cli is slated for deprecation, so I left it. It will need migration or removal before cli consumes the renamed `@x402r/evm` (which only exposes the kebab `auth-capture/*` subpaths).

A1igator:
Independently hit this during a Base Sepolia e2e of the garbage-detector arbiter example, and this fix resolves it.

**Repro of the bug (before this PR):** with `@x402r/evm@0.2.0-alpha.1` (scheme = `auth-capture`), every settlement made the merchant serve 200 and the facilitator settle on-chain (`success:true`), but the `requirements.scheme !== 'authCapture'` gate early-returned on all of them — the arbiter's `/verify` was never called, verdict count never moved, no capture/void ever fired. Silent, as described.

**After gating on `AUTH_CAPTURE_SCHEME` (= `"auth-capture"`):** the merchant→arbiter forward fires, the arbiter evaluates (PASS on `/weather`, FAIL on `/garbage`), and the PASS path executes an on-chain capture (escrow → merchant), e.g. tx `0x169d394e0c0618bc8548ea64491cb0c3f5743d4829b191112d5eb66d770facae`.

Confirms the gate constant is the right fix. One adjacent note from the same run (not this PR's scope): consumers also need to read the `paymentInfoWire` POST field rather than `paymentPayload` to reconstruct `PaymentInfo` — the example arbiter was still on the old shape; fixed separately on the examples side.

A1igator:
### SDK Review — Round 3

No new blocking issues since commit `25606da` (axes: tests, conventions, dead code, example alignment). Changes since round 2 are the two prior doc fixes plus routine dependency bumps; the `forwardToArbiter` logic and tests are unchanged and were cleared in rounds 1–2.

- Doubled `EIP3009_TOKEN_COLLECTOR_ADDRESS (EIP3009_TOKEN_COLLECTOR_ADDRESS)` parenthetical in `http-wire-capture.ts` — fixed.
- `README.md` camelCase `@x402r/evm/authCapture/{server,facilitator,client}` subpaths — migrated to kebab `auth-capture/`, client correctly pointed at `@x402/evm/auth-capture/client`.

**Still open (both sub-threshold, optional):**
- `forward-to-arbiter.test.ts:178` value-canary assertion (`expect(AUTH_CAPTURE_SCHEME).toBe('auth-capture')`) — the behavioral spy assertion below it is the real gate.
- `viem` / `@x402/evm` listed as `@x402r/helpers` peers without a direct `src/` import (transitively required via `@x402r/evm`, so defensible).

**New (sub-threshold note, not blocking):**
- This PR bumps `viem` to `^2.50.3` in `examples`, `cli`, `core`, and `sdk`, but leaves `packages/helpers/package.json` (the package being modified) at `^2.48.11` in both peer and dev. The lockfile resolves all of them to a single `viem@2.52.0`, so there is no resolution split — only a declared-floor inconsistency in the target package.

---

A1igator (approved):
Round 3 clean — prior-round issues resolved, remaining notes are sub-threshold/optional. forwardToArbiter gate + scheme-rename alignment verified across rounds. LGTM.

Issue #186, opened 2026-06-02
Adopt isolatedDeclarations for faster .d.ts generation (follow-up to tsdown migration)


vraspar:
Follow-up to the tsup → tsdown build migration.

tsdown generates declarations via rolldown-plugin-dts. By default (no `isolatedDeclarations` in tsconfig) it uses the TypeScript compiler — reliable but slower. With `isolatedDeclarations: true` in tsconfig it switches to the oxc-transform fast path, making declaration emit a near-instant syntax-only operation.

The tsdown migration intentionally did **not** enable it, to keep that PR a pure build-tool swap with an unchanged source surface. This issue tracks adopting it separately.

## Measured cost

Via `tsc -p tsconfig.build.json --noEmit --isolatedDeclarations` on the migrated branch:

| Package | Errors | Where |
|---|---|---|
| `@x402r/core` | 19 | `src/config/index.ts` — address constants need an explicit `: Address` annotation instead of `as const satisfies Address`, plus a few expression-inference cases |
| `@x402r/sdk` | 6 | `src/plugins/index.ts` — plugin factory functions need explicit return types |
| `@x402r/cli` | 0 | — |
| `@x402r/helpers` | 0 | — |

Test files add zero additional errors. So ~25 explicit annotations across two files.

## Plan

1. Enable `isolatedDeclarations: true` in `tsconfig.base.json`.
2. Add the explicit annotations until `tsc --isolatedDeclarations` is clean.
3. Verify the `.d.ts` surface is unchanged — note the annotations deliberately make some inferred public types explicit (e.g. address constants become `Address` rather than the literal type).

PR #187, opened 2026-06-02
build: migrate from tsup to tsdown

vraspar:
## Why

tsup is in maintenance mode and its frozen DTS toolchain breaks on TypeScript 6: the blocked dependabot bump #154 (`typescript 5.9.3 → 6.0.3`) fails the tsup DTS build with `error TS5101: Option 'baseUrl' is deprecated` — `baseUrl` is injected by tsup's own dts machinery (no project tsconfig sets it). tsdown declares a `^5 || ^6` TypeScript peer and uses a different dts engine, so it unblocks TS 6.

This migrates the four publishable packages (`@x402r/core`, `sdk`, `cli`, `helpers`) to tsdown while keeping the **published runtime + type surface unchanged**. TypeScript stays at 5.9.3 here — once this merges, #154 re-runs green and lands separately.

## Commit 1 = vanilla codemod, isolated

`df86fc8` is the raw output of `pnpm dlx tsdown-migrate . packages/*` and nothing else, so the tool's mechanical changes are reviewable separately from the hand changes that follow. It's committed `--no-verify` on purpose: the codemod defaults ESM output to `.mjs`/`.d.mts`, which mismatches the `.js`/`.d.ts` `exports` maps and fails typecheck — fixed in the very next commit.

## Beyond the vanilla codemod

- **Restore `.js`/`.d.ts` surface** (`fixedExtension: false`) — tsdown defaults to `.mjs`/`.d.mts` for `platform: node`; this keeps the exact tsup filenames so the existing `exports` maps resolve unchanged.
- **Explicit `target: 'node22'`** — runtime support is decoupled from tsdown's own Node ≥ 22.18 *build* requirement.
- **Node floor, build tooling only** — `.nvmrc` and root (private) `engines` → `>=22.18.0`. **No `engines` added to any published package** (that would wrongly constrain consumers' runtime).
- **Native publint + attw gate** — run in the tsdown build (`level: error`, `profile: esm-only`) for **all four packages, including `@x402r/cli`**, which the old root scripts skipped. `test:pkg` is now the explicit `turbo run build --force` gate, wired into CI + release. Removed the standalone `publint`/`attw` scripts, the four `prepublishOnly` hooks, and the inline CI steps; swapped `@arethetypeswrong/cli` → `@arethetypeswrong/core` (what tsdown uses).
- **size-limit recalibrated** — `@x402r/core` `config.js` and `@x402r/helpers` `index.js` 2 → 3 kB. Investigated: the growth is preserved JSDoc comments + small-entry inlining (tsup re-exported from a shared chunk and stripped comments), **not functional bloat** — total dist JS is roughly flat.
- **Pinned tsdown `0.22.1`** (pre-1.0).

## Kept manual / deferred (deliberate)

- **`exports` maps kept hand-written.** Evaluated `exports: true`; attw passed across all resolution modes, but the generated shape isn't byte-equivalent (flat strings, dropped per-subpath `types`, added `./package.json`). Per a surface-stability gate, kept the explicit maps.
- **`isolatedDeclarations` deferred** → tracked in #186 (measured ~25 annotations across two files; this PR makes **zero** source changes).

## Verification

- `build` / `typecheck` (6/6) / 216 unit tests / `knip` / `biome` / `size-limit` all green.
- **`.d.ts` export surface identical** to the tsup baseline across all 13 entries (verified via the TypeScript compiler API).
- attw + publint clean for all four; `x402r --help` runs (shebang + executable bit preserved); `__VERSION__` still injected at build.
- **TS6 unblock proven**: under `typescript@6.0.3`, both `build` and `typecheck` pass — the same build that fails on tsup. Reverted; the bump stays in #154.

## Note

This branch was cut before #184 (a production-deps bump) landed on `main` and does not touch `examples/`; a lockfile merge may be needed at merge time (resolve via merge-from-main, not rebase).


A1igator (dismissed):

vraspar:
### SDK Review

Found 1 issue (axes: package surface, migration config):

1. **[Package Surface]** `target: 'node22'` raises the emitted runtime baseline without a published runtime contract. The PR says the consumer runtime surface is unchanged and correctly avoids adding `engines` to the published packages, but every tsdown config now targets Node 22 while the existing compiler target is ES2022. That means future output can preserve Node 22-only syntax without package metadata telling consumers, which is a quiet compatibility change from the current ES2022 build target. Prefer `target: 'es2022'` to match `tsconfig.base.json`, or publish an explicit runtime floor if Node 22 is now required.

   https://github.com/BackTrackCo/x402r-sdk/blob/6ec0d06db438f02e8a75d5c3007e9687964301ce/packages/core/tsdown.config.ts#L23

---
Generated with [Codex](https://Codex.ai/code) using review-sdk skill

vraspar:
### SDK Review (follow-up)

The `target: 'node22'` baseline note is already covered in the earlier review comment; these are additional findings from a build-and-source pass on `6ec0d06` (built all four packages; read tsdown + publint sources to check gate semantics).

1. **[Release gate — concrete] The publint check is now weaker than the old `--strict` gate.** The previous gate ran `publint --strict`, which elevates warnings to errors and fails on them. The tsdown config instead sets `publint: { level: 'error' }` with no `strict`. Per publint's own API, `level: 'error'` *filters the result down to error-level messages only* (dropping warnings and suggestions), and `strict: true` is the separate option that reports warnings as errors. tsdown's integration also only fails on warnings when the global `failOnWarn` is set, which defaults to `false`. Net effect: **publint warnings will no longer fail CI or release** — the opposite of the prior `--strict` behavior. It is green today only because there are zero warnings. Suggest adding `strict: true` alongside `level: 'error'` in each config (per publint docs, strict-elevated warnings still surface even with `level: 'error'`).

   https://github.com/BackTrackCo/x402r-sdk/blob/6ec0d06db438f02e8a75d5c3007e9687964301ce/packages/core/tsdown.config.ts#L25

2. **[Defense-in-depth] Removing the four `prepublishOnly` hooks drops the last-line publish guard.** Those hooks ran publint/attw at actual `publish` time regardless of how publish was invoked; validation now lives only in the release workflow's `test:pkg` step. That covers the normal CI path, but a manual or out-of-band `pnpm publish` would have no validation gate. Fine if the team only ever publishes via the workflow — worth confirming that's intentional rather than incidental to the migration.

Two notes (not blockers):
- **attw rule handling** changed from `--ignore-rules no-resolution cjs-resolves-to-esm` to `profile: 'esm-only'`. For an ESM-only package this is the principled equivalent and arguably stricter (`no-resolution` is no longer suppressed); attw passes clean, so no regression.
- **Published bytes** grew ~1 kB on `core/config.js` and `helpers/index.js` (the 2→3 kB size-limit bumps) because tsdown preserves JSDoc in emitted JS where tsup stripped it. API/types are unchanged, but the `.js` output is not literally byte-identical — the "surface unchanged" claim holds for API/types, not raw bytes.

Verified holding up: build green for all four, publint + attw clean (now including `@x402r/cli`), `.js`/`.d.ts` filenames preserved, `exports` maps untouched, CLI bin shebang + exec bit intact.

vraspar:
Thanks for the review — addressed in `b90150d` (then merged `main`).

**1. `target: 'node22'` → `'es2022'`** ✅ Agreed. Matches `tsconfig.base.json`'s compiler target and avoids coupling emitted syntax to a Node version. Output is byte-identical today (the source is ES2022-capped by tsconfig, so nothing newer can leak) — a zero-risk consistency fix, and closer to tsup's behavior (esbuild took its level from tsconfig).

**2. publint `{ level: 'error' }` → `{ strict: true }`** ✅ Agreed, good catch — this was a real regression. Confirmed against tsdown's source: it spreads the config straight into `publint()`, and publint's `level: 'error'` filters results to error-type only (warnings dropped before tsdown sees them), whereas the old `--strict` *promoted* warnings to errors so they failed. `strict: true` restores that. Verified there are zero current warnings (`strict: true` surfaces nothing extra), so the fix is risk-free today and re-arms the gate going forward. (attw is unaffected — `level: 'error'` there genuinely fails per the same source; only publint has the level/strict split.)

**3. Removed `prepublishOnly` / last-line publish guard** — pushing back; leaving as-is, and I believe it's intentional rather than incidental:
- Packages publish *only* via `release.yml` with npm Trusted Publishing (OIDC + provenance). A manual `pnpm publish` can't mint the OIDC token, so an out-of-band publish isn't a real path.
- The workflow path is gated — `test:pkg` builds + runs native publint/attw across all four right before `changeset publish`.
- Restoring `prepublishOnly` would mean re-adding the standalone publint/attw CLIs (+ `@arethetypeswrong/cli`) this PR intentionally dropped for the native gate — so it works against that decision.

If we want belt-and-suspenders regardless, the native-consistent option is a per-package `prepublishOnly: "tsdown --publint --attw"` (no standalone CLIs). Happy to add it, but I'd lean against — a redundant publish-time rebuild guarding a path that can't happen.

**Notes:** attw `profile: 'esm-only'` kept (agreed it's the principled/stricter equivalent). Changeset wording tightened: API + type declarations are unchanged, but the emitted `.js` isn't byte-identical (Rolldown retains JSDoc that esbuild stripped).

A1igator (approved):

PR #188, opened 2026-06-02
docs(sdk): scrub upstream-org cross-refs from published source comments

vraspar:
Removes the forbidden `x402-foundation/x402#…` cross-reference strings from `@x402r/sdk` source comments (they create sticky timeline events in the upstream repo) and rewrites them as plain prose that preserves the technical meaning.

- `sdk/src/types.ts` and `sdk/src/actions/erc8004-helpers.ts`: the `#931` / `#1024` refs (both about the same upstream x402 reputation extension) → "the upstream x402 reputation extension spec / discussion".
- Comment-only — no code, type, or behavior change; `tsc` clean. Diffstat: 2 files, +6/-6.

Per the repo convention of never embedding `org/repo#N` to non-BackTrackCo orgs in committed artifacts. No changeset (no user-facing change).

### DIFF
--- a/packages/sdk/src/actions/erc8004-helpers.ts
+++ b/packages/sdk/src/actions/erc8004-helpers.ts
@@ -54,15 +54,15 @@ export function extractArbiterIdentity(
 /**
  * Extract agent registrations from the upstream x402 reputation extension.
  *
- * Parses `extensions["reputation"].info.registrations` per the upstream spec
- * (x402-foundation/x402 PR #1024). Returns validated registrations as an array.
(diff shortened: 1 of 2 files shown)


vraspar:
Closing — the premise is faulty. The `org/repo#N` strings here were in source-file comments, which don't create GitHub cross-references (those come from commit messages, PR/issue bodies, comments, and branch names — not blob contents), so they never pinged upstream. The original references were useful pointers to the upstream reputation-extension spec/discussion; the prose rewrite is strictly less informative. No change needed.

PR #189, opened 2026-06-02
feat(core)!: add erc3009+permit2 collectors to chain config; remove tokenCollector

vraspar:
## What

Adds `collectors: { erc3009, permit2 }` to `X402rChainConfig` so callers can resolve the canonical `commerce-payments v1.0.0` collector by asset transfer method (`erc3009` for `ReceiveWithAuthorization` payloads, `permit2` for Permit2 payloads), and exports the `CollectorAddresses` type.

**Removes** the `tokenCollector` chain-config field, the top-level `tokenCollector` const, and its `@x402r/helpers` re-export. There is **no `@deprecated` alias** — in the 0.x/alpha line we remove and migrate rather than carry a back-compat shim.

## Breaking change

- `X402rChainConfig.tokenCollector` → use `.collectors.erc3009` (ERC-3009) or `.collectors.permit2` (Permit2).
- The top-level `tokenCollector` export is removed from both `@x402r/core` and `@x402r/helpers`.

Internal `@x402r/core` config readers (`signReceiveAuthorization` and the ERC-3009 test/fork helpers) were migrated to `getChainConfig(id).collectors.erc3009`. The unrelated `tokenCollector` function-parameter / `collectorData` / ABI identifiers are untouched.

## Validation

- `pnpm --filter @x402r/core test config` — 36 passed
- `pnpm build` + `pnpm typecheck` — green across all packages (incl. `@x402r/helpers`, whose re-export was removed)
- Rebased onto current `main` (post-tsdown migration); builds clean under the new tooling.

### DIFF
--- a/.changeset/core-config-collectors.md
+++ b/.changeset/core-config-collectors.md
@@ -1,0 +1,6 @@
+---
+"@x402r/core": minor
+"@x402r/helpers": minor
+---
+
+Add `collectors` (`{ eip3009, permit2 }`) and `getCollectorAddress(chainId, method)` to resolve a payment collector by transfer method (re-exported from `@x402r/helpers`). Remove `tokenCollector` — use `collectors.eip3009` or `getCollectorAddress(id, 'eip3009')`.
--- a/.changeset/core-config-collectors.md
+++ b/.changeset/core-config-collectors.md
@@ -3,4 +3,4 @@
 "@x402r/helpers": minor
 ---
 
-Add `collectors` (`{ eip3009, permit2 }`) and `getCollectorAddress(chainId, method)` to resolve a payment collector by transfer method (re-exported from `@x402r/helpers`). Remove `tokenCollector` — use `collectors.eip3009` or `getCollectorAddress(id, 'eip3009')`.
+Add `collectors: { eip3009, permit2 }` to chain config to resolve the canonical commerce-payments collector by transfer method, and remove the stale single-valued `tokenCollector` field + const. Migrate `tokenCollector` -> `collectors.eip3009` (or `.permit2`).
--- a/packages/core/src/config/index.ts
+++ b/packages/core/src/config/index.ts
@@ -31,11 +31,22 @@ export interface HookSingletonAddresses {
   paymentIndexRecorderHook: Address
 }
 
+/**
+ * Canonical `commerce-payments v1.0.0` payment collectors, keyed by the
+ * `assetTransferMethod` a merchant advertises. Pick `eip3009` for
+ * `ReceiveWithAuthorization` payloads and `permit2` for Permit2 payloads.
+ */
+export interface CollectorAddresses {
+  eip3009: Address
+  permit2: Address
+}
+
 export interface X402rChainConfig {
   name: string
   chainId: number
   authCaptureEscrow: Address
-  tokenCollector: Address
+  /** Payment collectors keyed by asset transfer method (`eip3009` / `permit2`). */
+  collectors: CollectorAddresses
   protocolFeeConfig: Address
   receiverRefundCollector: Address
   usdc: Address
@@ -70,13 +81,6 @@ export interface X402rChainConfig {
 /** AuthCaptureEscrow at the canonical `commerce-payments at v1.0.0` deployment. */
 export const authCaptureEscrow =
   '0xBdEA0D1bcC5966192B070Fdf62aB4EF5b4420cff' as const satisfies Address
-
-/**
- * Primary token collector. Currently aliases the canonical
- * `ERC3009PaymentCollector` — Permit2 lands in a follow-up PR.
- */
-export const tokenCollector =
-  '0x0E3dF9510de65469C4518D7843919c0b8C7A7757' as const satisfies Address
 
 export const protocolFeeConfig =
   '0xBe2d24614F339a1eB103A399F93AA2a39Ca815Bc' as const satisfies Address
@@ -163,13 +167,16 @@ export const commercePaymentsAddresses = {
 
 const PROTOCOL_ADDRESSES = {
   authCaptureEscrow,
-  tokenCollector,
+  collectors: {
+    eip3009: commercePaymentsErc3009PaymentCollector,
+    permit2: commercePaymentsPermit2PaymentCollector,
+  },
   protocolFeeConfig,
   receiverRefundCollector,
   factories,
   conditions,
   hooks,
-} as const
+} as const satisfies Omit<X402rChainConfig, 'name' | 'chainId' | 'usdc'>
 
 /** Build a chain config by spreading unified protocol addresses + chain-specific USDC */
 function chainConfig(
@@ -251,6 +258,23 @@ export function getFactoryAddress(
   if (address === zeroAddress) {
     throw new ConfigError(
       `${factory} factory is not deployed on ${config.name}`,
+    )
+  }
+  return address
+}
+
+export function getCollectorAddress(
+  chainId: number,
+  method: keyof CollectorAddresses,
+): Address {
+  const config = getChainConfig(chainId)
+  if (!config.collectors) {
+    throw new ConfigError(`Collectors are not deployed on ${config.name}`)
+  }
+  const address = config.collectors[method]
+  if (address === zeroAddress) {
+    throw new ConfigError(
+      `${method} collector is not deployed on ${config.name}`,
     )
   }
   return address
--- a/packages/core/src/config/index.ts
+++ b/packages/core/src/config/index.ts
@@ -263,23 +263,6 @@ export function getFactoryAddress(
   return address
 }
 
-export function getCollectorAddress(
-  chainId: number,
-  method: keyof CollectorAddresses,
-): Address {
-  const config = getChainConfig(chainId)
-  if (!config.collectors) {
-    throw new ConfigError(`Collectors are not deployed on ${config.name}`)
-  }
-  const address = config.collectors[method]
-  if (address === zeroAddress) {
-    throw new ConfigError(
-      `${method} collector is not deployed on ${config.name}`,
-    )
-  }
-  return address
-}
-
 export function getConditionSingletons(
   chainId: number,
 ): ConditionSingletonAddresses {
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -205,6 +205,7 @@ export {
 // Config
 // ---------------------------------------------------------------------------
 export type {
+  CollectorAddresses,
   ConditionSingletonAddresses,
   FactoryAddresses,
   HookSingletonAddresses,
@@ -217,6 +218,7 @@ export {
   factories,
   fromNetworkId,
   getChainConfig,
+  getCollectorAddress,
   getConditionSingletons,
   getFactoryAddress,
   getFactoryAddresses,
@@ -229,7 +231,6 @@ export {
   protocolFeeConfig,
   receiverRefundCollector,
   supportedChainIds,
-  tokenCollector,
   toNetworkId,
   x402rChains,
 } from './config/index.js'
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -218,7 +218,6 @@ export {
   factories,
   fromNetworkId,
   getChainConfig,
-  getCollectorAddress,
   getConditionSingletons,
   getFactoryAddress,
   getFactoryAddresses,
--- a/packages/core/src/payment/erc3009.ts
+++ b/packages/core/src/payment/erc3009.ts
@@ -65,7 +65,8 @@ export async function signReceiveAuthorization(
 ): Promise<SignReceiveAuthorizationReturnType> {
   const { account, chainId, paymentInfo } = parameters
   const chainConfig = getChainConfig(chainId)
-  const tokenCollector = parameters.tokenCollector ?? chainConfig.tokenCollector
+  const tokenCollector =
+    parameters.tokenCollector ?? chainConfig.collectors.eip3009
   const escrowAddress =
     parameters.escrowAddress ?? chainConfig.authCaptureEscrow
   const nonce = computeEscrowNonce(chainId, escrowAddress, paymentInfo)
--- a/packages/core/src/payment/permit2.ts
+++ b/packages/core/src/payment/permit2.ts
@@ -1,10 +1,7 @@
 import type { Address, Hex } from 'viem'
 import { encodeFunctionData, erc20Abi, getAddress } from 'viem'
 import type { LocalAccount } from 'viem/accounts'
-import {
-  commercePaymentsPermit2PaymentCollector,
-  getChainConfig,
-} from '../config/index.js'
+import { getChainConfig } from '../config/index.js'
 import type { PaymentInfo } from '../types/index.js'
 import { computeEscrowNonce } from './hashing.js'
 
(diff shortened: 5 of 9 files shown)


vraspar:
### SDK Review

Found 3 issues:

1. **[API]** Rename the public collector key from `erc3009` to `eip3009`. The config says collectors are keyed by `assetTransferMethod`, and the actual wire value is `eip3009 | permit2`. With `collectors.erc3009`, consumers cannot do `config.collectors[assetTransferMethod]` without a translation table.

   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/packages/core/src/config/index.ts#L34-L49

2. **[API]** Make collector support method-specific instead of requiring both methods on every chain. Future chains may only have one usable path, because EIP-3009 depends on token support, Permit2 depends on chain deployment/allowance flow, and x402r still needs the matching auth-capture collector deployed. Prefer `collectors: Partial<Record<'eip3009' | 'permit2', Address>>` plus `getCollectorAddress(chainId, method)` that throws `ConfigError` when the requested method is unavailable. Do not use zero addresses or fallback between methods.

   Also use this resolver in both signing helpers: `signReceiveAuthorization` for `eip3009`, and `signPermit2Authorization` for `permit2`. Keep the existing `tokenCollector` parameter as an explicit override, and re-export the resolver from `@x402r/helpers`.

   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/packages/core/src/config/index.ts#L167-L172
   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/packages/core/src/payment/erc3009.ts#L68-L69
   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/packages/core/src/payment/permit2.ts#L115-L118
   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/packages/helpers/src/index.ts#L4-L12

3. **[Tests / changeset]** Update the tests and release note for the new shape. The unit test currently asserts every supported chain has both collectors, which would reject a valid future one-method chain. Test that Base returns both methods today, that the resolver throws for a missing method, and that the fork smoke test checks each configured collector entry. The changeset should point users to `getCollectorAddress(id, 'eip3009' | 'permit2')` or the `eip3009` key, not `.collectors.erc3009`.

   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/packages/core/tests/config.test.ts#L99-L107
   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/packages/core/tests/integration/config-addresses.fork.test.ts#L82-L88
   https://github.com/BackTrackCo/x402r-sdk/blob/35ba64f05c1fee25523742c761a43bd6c7f4368e/.changeset/core-config-collectors.md#L6

vraspar:
Thanks for the review. Status on the three findings:

Accepted/done:
- eip3009 key: the collector key is eip3009 (not erc3009) across the config type, the chain literal, signReceiveAuthorization's default, the unit + fork tests, and the erc3009-helper test setup — so config.collectors[assetTransferMethod] works directly. The review was against an earlier SHA; this landed in the current commit.
- getCollectorAddress resolver: added in core (throws ConfigError for an unsupported chain and for a zero-address entry) and re-exported from @x402r/helpers.
- Changeset + tests: the changeset now points users to collectors.eip3009 / getCollectorAddress(id, 'eip3009'); added a getCollectorAddress test block; the fork smoke test verifies both the eip3009 and permit2 collector entries have bytecode.

Pushbacks:
- Total Record, not Partial: the collectors stay { eip3009, permit2 } (both-or-neither per chain) because they're chain-invariant canonical commerce-payments constants — the real eip3009-vs-permit2 split is per-token (bridged USDC.e has no eip3009) and is handled by assetTransferMethod, not the chain map. getCollectorAddress abstracts access so moving to Partial later is non-breaking, and the "resolver throws for a missing method" test is moot while the map is Total.
- No resolver wired into the signers: signReceiveAuthorization / signPermit2Authorization are being removed at 0.3.0 (payer signing moves to upstream @x402/evm), so I'm not threading new logic into code that's about to be deleted; erc3009.ts gets only the key rename and permit2.ts keeps its bare const.

Validated: @x402r/core config tests 39 passed; pre-commit build + typecheck green across all packages.

vraspar:
Slimmed in `daba7e7`: removed `getCollectorAddress` (the resolver) and its `@x402r/core` / `@x402r/helpers` re-exports + self-test.

**Why:** it had no consumers. Every flow already gets the collector elsewhere — the payer signers return it; the CLI and the upstream `@x402/evm` client embed their own; the example scenarios destructure it from the signer's return value; and the one manual spot (arbiter-examples) imports the singular `tokenCollector` const (which this PR removes — its migration target is `collectors.eip3009`, member access). A whole-workspace grep found zero `getCollectorAddress` usage. After the signers are removed at 0.3.0, a merchant/arbiter calling `authorize`/`charge`/`refund` can resolve the collector via `getChainConfig(id).collectors.{eip3009,permit2}` or the scheme's exported constant — the resolver's only extra value was a throw-on-missing, and collectors are never zero in the registry. If a guarded resolver is ever wanted (e.g. select-by-`assetTransferMethod`), it's ~12 lines reintroduced *with* its consumer, so we're not shipping consumer-less public API to carry to 1.0.

**Kept:** the `collectors: { eip3009, permit2 }` chain-config data + `CollectorAddresses` type — `signReceiveAuthorization` reads `collectors.eip3009` today, and it's the migration target for the removed `tokenCollector`. Also repointed `permit2.ts` to `collectors.permit2` for consistency with `erc3009.ts`. The stale single-valued `tokenCollector` removal stands.

tofu-miso (commented):
## Review — `feat(core)!: add erc3009+permit2 collectors to chain config; remove tokenCollector`

Reviewed the full diff against `main`, the surrounding `config/index.ts`, the signer call-sites, the tests, the examples tree, and the three review/response threads. Net assessment: **this is a clean, well-scoped change and is correct.** The collector-key naming matches the wire spec, every internal reader is migrated, and there are no dangling references to the removed symbols. A few small nits and one mild test-coverage regression below; none are blockers.

### Summary of what the diff does
- Adds `CollectorAddresses { eip3009; permit2 }` and replaces the single `X402rChainConfig.tokenCollector: Address` field with `collectors: CollectorAddresses`.
- Removes the top-level `tokenCollector` const and its `@x402r/core` + `@x402r/helpers` re-exports.
- Migrates the two internal readers: `signReceiveAuthorization` → `collectors.eip3009`, `signPermit2Authorization` → `collectors.permit2`, plus the erc3009 test helper.
- Tightens `PROTOCOL_ADDRESSES` with `satisfies Omit<X402rChainConfig, 'name' | 'chainId' | 'usdc'>`.

### Correctness — no bugs found
- **Value preservation verified.** Old `tokenCollector = 0x0E3d…7757` is identical to `commercePaymentsErc3009PaymentCollector` (`collectors.eip3009`), and `signPermit2Authorization`'s old default `commercePaymentsPermit2PaymentCollector` is identical to `collectors.permit2` (`0x9924…aB26`). So neither signer changes behavior — confirmed by the value pins still passing in `erc3009.test.ts:88` and `permit2.test.ts:98`.
- **Key naming is right.** The wire value is `'eip3009' | 'permit2'` (`helpers/src/x402r-defaults.ts:69`, `cli/src/bin.ts:19`), so `eip3009` (not `erc3009`) is the correct key and `config.collectors[assetTransferMethod]` resolves directly. Good catch carried over from review #1.
- **No dangling references.** Grepped the whole workspace: every remaining `tokenCollector` is a *function parameter / return value* on `signReceive…`, `signPermit2…`, `authorize`, `charge` (intentionally untouched) or the unrelated `EIP3009_TOKEN_COLLECTOR` from `@x402r/evm`. Nothing still imports the removed const or reads `.tokenCollector` off a config. The migration is complete.
- The `satisfies Omit<…>` on `PROTOCOL_ADDRESSES` is a nice touch — it now structurally guarantees the spread covers every non-chain-specific field.

### Nits
1. **Mild test-coverage regression (`config.test.ts:99-107`).** The old test pinned the collector to a named constant (`expect(config.tokenCollector).toBe(tokenCollector)`). The replacement only regex-checks the format and asserts the two differ. Since `collectors` is spread from `PROTOCOL_ADDRESSES` into every chain, the loop also re-checks the *same* two constants N times. Consider pinning the exact canonical values instead, e.g. `expect(config.collectors.eip3009).toBe(commercePaymentsErc3009PaymentCollector)` / `…permit2).toBe(commercePaymentsPermit2PaymentCollector)`. That restores the regression guard the old assertion provided. (Exact values are still pinned indirectly via the signer tests, so this is low-severity.)
2. **Changeset bump (`minor`) vs. `!` in the title.** Removing a public export is breaking; the `minor` bump relies on the 0.x convention (which the PR body states explicitly), so it's defensible — just flagging the mixed signal between the conventional-commit `!` and the `minor` changeset for anyone scanning.

### Comment quality
- The `CollectorAddresses` doc block (`config/index.ts:34-42`) is accurate and genuinely useful — explains the `eip3009`/`permit2` ↔ payload mapping.
- The field-level `/** Payment collectors keyed by asset transfer method… */` on `collectors:` (`:48`) largely restates the interface doc directly above it. Harmless for IntelliSense, but slightly redundant.
- `// eip3009 collector and permit2 collector are distinct canonical addresses` (`config.test.ts:106`) just restates the assertion on the next line — could drop it.
- Good cleanup: the verbose Permit2 `tokenCollector?` param doc in `permit2.ts` was correctly trimmed to "default: from chain config" now that it reads from config rather than a bare const.

### Author's last comment assessment
The last comment (commit `daba7e7`) removes the just-added `getCollectorAddress` resolver as consumer-less. The core conclusion holds — I confirmed a whole-workspace grep finds **zero** `getCollectorAddress` usages, and the final diff contains no resolver, so the PR's end state is internally coherent.

Two caveats on the reasoning:
- The supporting claim that "the one manual spot (arbiter-examples) imports the singular `tokenCollector` const" doesn't check out — `examples/arbiter/` (`approve-refund.ts`, `distribute-fees.ts`, `review-evidence.ts`) has **no** collector or `tokenCollector` reference at all. The conclusion (no live consumers) is still correct; that particular justification is just inaccurate.
- Slight internal tension across the thread: comment #2 justified keeping a Total `Record` by saying "`getCollectorAddress` abstracts access so moving to `Partial` later is non-breaking," then comment #3 removed that very abstraction. With direct member access (`config.collectors.eip3009`) as the only path now, a future switch to `Partial<Record<…>>` *would* be a type-breaking change (the field becomes `Address | undefined`). Worth keeping in mind if the one-method-chain scenario ever materializes.

Net: removing consumer-less public API before 1.0 is the right call, and the Total-`Record` rationale (collectors are chain-invariant canonical constants; the real eip3009-vs-permit2 split is per-token via `assetTransferMethod`) is sound. The pushback on not wiring a resolver into signers that are slated for removal at 0.3.0 is also reasonable, modulo trusting that roadmap.

**LGTM** once you decide on the `config.test.ts` value-pin nit.

vraspar:
Thanks @tofu-miso — appreciate the re-verify. Responses:

- **Exact-value pin (test nit):** the canonical collector values are already pinned by the signer tests — `erc3009.test.ts` asserts the resolved collector equals the canonical ERC-3009 address and `permit2.test.ts` does the same for Permit2 — so the suite isn't missing that guard; the `collectors` test deliberately covers shape + distinctness. (Also the literal `import { commercePaymentsErc3009PaymentCollector } from '../src/index.js'` wouldn't compile — those consts aren't barrel-exported; it'd need a deep import.) Happy to add an exact-value assert in the collectors test via a deep import if you'd rather have it co-located, but it's redundant with the signer tests.
- **`minor` vs `!`:** intentional — in the 0.x line a `minor` *is* the breaking bump; the `!` is just the conventional-commit marker. Leaving it.
- **arbiter-examples (re my earlier comment):** small clarification — I meant the **separate `arbiter-examples` repo** (the `x402r-ai-garbage-detector` example), which imports the singular `tokenCollector` const from `@x402r/helpers`; the in-repo `examples/arbiter/` is indeed clean. Either way, agreed the resolver had no live consumer.
- **Total → Partial type-break:** correct — with member access, switching the field to `Partial` later would be type-breaking. That's the accepted tradeoff (no one-method chain exists today); if one ever does, we reintroduce a guarded resolver *with* its consumer rather than pre-shipping a consumer-less one.

Net: I think this is good to merge once it has an approving review.

A1igator:
### SDK Review — Round 2

Found 0 new issues since commit 35ba64f (axes: tests, conventions, dead code, example alignment).

- Collector key renamed `erc3009` to `eip3009`; `config.collectors[assetTransferMethod]` resolves directly. Verified across the config type, chain literal, both signer defaults, unit + fork tests, and the erc3009 test helper.
- Tests and changeset updated for the new shape; the fork smoke test covers both collector entries.
- `getCollectorAddress`: added in fcb2568, removed in daba7e7 as consumer-less with documented rationale. The add-then-remove left zero orphans: no references remain in source, tests, READMEs, or changesets, and the resolver's inline `ConfigError` throw sites are fully gone.

**Still open:** none.

**New:** none.

Verification notes:
- Both signer tests pin the canonical collector values through the new config-default path: `erc3009.test.ts` L80-88 calls `signReceiveAuthorization` with no override and asserts `0x0E3dF9510de65469C4518D7843919c0b8C7A7757`; `permit2.test.ts` L90-98 asserts `0x992476B9Ee81d52a5BdA0622C333938D0Af0aB26`. The `config.tokenCollector` pin removed from `config.test.ts` is covered by these.
- Every remaining `tokenCollector` identifier in `packages/core`, `packages/sdk`, `examples/`, and READMEs is a signer param/return name or ABI struct field, not a read of the removed config field. No example or barrel imports the removed const.
- `CollectorAddresses` is a load-bearing type export: `X402rChainConfig.collectors` references it, so declaration-emit requires it in the barrel.

---

tofu-miso:
Re-verified your responses — all four points resolved, accepting the pushback on each.

- **Test value-pin:** You're right, and I confirmed the compile point myself: `config.test.ts` barrel-imports from `../src/index.js`, and neither `commercePaymentsErc3009PaymentCollector` nor `…Permit2…` is re-exported from the core barrel (only via deep `./config/index.js`), so my suggested literal wouldn't have compiled. And the canonical values are already pinned through the signer tests — `erc3009.test.ts:88` asserts the config-resolved `collectors.eip3009` equals `0x0E3d…7757` and `permit2.test.ts:98` does the same for `0x9924…aB26`, both reading straight from chain config with no override. Since `collectors` is chain-invariant (spread from `PROTOCOL_ADDRESSES`), that's a real value guard. The shape+distinctness scope of the collectors test is fine as-is — no need for the redundant co-located assert.
- **`minor` vs `!`:** Accepted — that was only ever a flag for scanners, not a request. 0.x `minor` = breaking, intentional.
- **arbiter-examples:** Thanks for the clarification — the separate `arbiter-examples` repo, not in-repo `examples/arbiter/` (which is clean). Either way the workspace grep confirmed zero live `getCollectorAddress` consumers, so removing it pre-1.0 is right.
- **Total→Partial:** Agreed and resolved — switching the field to `Partial` later would be type-breaking, accepted as the tradeoff while no one-method chain exists; reintroduce a guarded resolver *with* its consumer if one ever does.

Nothing outstanding from my side. LGTM — good to merge.

A1igator (approved):

PR #190, opened 2026-06-02
fix(helpers): x402rDefaults emits relative deadline offsets by default (closes #183)

### LINKED ISSUE #183: x402rDefaults emits absolute deadlines frozen at call time; should default to relative offsets
## Summary

`x402rDefaults` (`packages/helpers/src/x402r-defaults.ts`) computes **absolute** `captureDeadline` / `refundDeadline` values at the moment it is called, rather than emitting the **relative** `captureDeadlineSeconds` / `refundDeadlineSeconds` offsets that the auth-capture server scheme is designed to resolve per request.

Because merchants typically call `x402rDefaults` once while building their route config at server startup, every payment requirement served for the lifetime of the process advertises the **same fixed deadline** computed at boot. The capture/refund window then shrinks with every passing request and eventually expires, instead of giving each authorization a fresh window.

## Where

`packages/helpers/src/x402r-defaults.ts`:

```ts
const DEFAULT_CAPTURE_WINDOW_SECONDS = 60 * 60 * 24      // 24 hours
const DEFAULT_REFUND_WINDOW_SECONDS  = 60 * 60 * 24 * 7  // 7 days

export function x402rDefaults(input: X402rDefaultsInput): AuthCaptureExtra {
  const nowSeconds = Math.floor(Date.now() / 1000)       // <-- evaluated once, at call time
  return {
    captureAuthorizer: input.captureAuthorizer,
    captureDeadline: input.captureDeadline ?? nowSeconds + DEFAULT_CAPTURE_WINDOW_SECONDS,
    refundDeadline:  input.refundDeadline  ?? nowSeconds + DEFAULT_REFUND_WINDOW_SECONDS,
    ...
  }
}
```

## Why this is wrong

The auth-capture server scheme already supports relative offsets and resolves them to absolute deadlines **per request** inside `enhancePaymentRequirements` (in the x402r-scheme `@x402r/evm` package). Its own merchant-field hints call this out:

> Set `extra.captureDeadlineSeconds` (relative, recommended) or `extra.captureDeadline` (absolute).

The canonical server example uses the relative form for exactly this reason:

```ts
extra: {
  captureAuthorizer,
  captureDeadlineSeconds: 30 * 86400,  // scheme converts to absolute per request → fresh window each time
  refundDeadlineSeconds:  60 * 86400,
  feeRecipient: zeroAddress,
  minFeeBps: 0,
  maxFeeBps: 100,
}
```

`x402rDefaults`, the quick-start helper, currently defaults to the *not*-recommended absolute form and freezes it at call time, so the helper steers merchants toward the footgun rather than away from it.

## Suggested fix

Have `x402rDefaults` emit the relative offsets by default (`captureDeadlineSeconds` / `refundDeadlineSeconds`) so the server scheme produces a fresh per-request deadline, and stop calling `Date.now()` in the helper. Continue to allow callers to pass absolute `captureDeadline` / `refundDeadline` explicitly when they really want a fixed wall-clock deadline.

This also removes the `Date.now()` call from the helper, which is a nice side benefit for testability/determinism.

## Context

Found while migrating the x402r landing-page integration snippets onto the auth-capture scheme. The snippets reference `x402rDefaults`; this is purely about the helper's default behavior, not the snippets.

Related: `x402rDefaults` is also not yet re-exported from `@x402r/helpers`' `index.ts` (separate follow-up).

vraspar:
Closes #183.

`x402rDefaults` computed **absolute** `captureDeadline` / `refundDeadline` from the wall clock at call time — so a merchant calling it once at server startup served a fixed, shrinking window for the lifetime of the process. It now emits **relative** `captureDeadlineSeconds` / `refundDeadlineSeconds` (defaults `86400` / `604800`) that the auth-capture server scheme resolves to absolute deadlines **per request** — a fresh window every time — and the helper no longer reads the wall clock (deterministic output).

- **Absolute deadlines are now opt-in:** pass both `captureDeadline` + `refundDeadline` to emit a fixed wall-clock window instead.
- Return type is a self-contained, helpers-local `X402rDefaultsExtra` (`Omit<AuthCaptureExtra, 'captureDeadline' | 'refundDeadline'> & (relative | absolute)`) — no new value import from `@x402/evm` / `@x402r/evm` (type-only).
- Tests cover relative-by-default, the absolute opt-in (both fields required together), and determinism (two calls deep-equal).

`minor` changeset for `@x402r/helpers`.

**Note:** `@x402r/evm`'s `AuthCaptureExtra` types only the absolute deadline fields (the server scheme reads the `*Seconds` offsets at runtime but doesn't type them). A small scheme-side type addition would let this output assign to `requirements.extra` without the local type — tracked as a separate follow-up.

### DIFF
--- a/.changeset/helpers-x402rdefaults-relative-offsets.md
+++ b/.changeset/helpers-x402rdefaults-relative-offsets.md
@@ -1,0 +1,5 @@
+---
+"@x402r/helpers": minor
+---
+
+`x402rDefaults` now emits relative deadline offsets (`captureDeadlineSeconds` / `refundDeadlineSeconds`, defaults 86400 / 604800) instead of absolute deadlines computed at call time; the auth-capture server scheme resolves each to an absolute deadline per request. Pass absolute `captureDeadline` / `refundDeadline` to fix either window — each deadline is resolved independently.
--- a/packages/helpers/src/x402r-defaults.ts
+++ b/packages/helpers/src/x402r-defaults.ts
@@ -175,6 +175,18 @@ export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
   // footgun the absolute defaults used to cause.
   let capture: { captureDeadline: number } | { captureDeadlineSeconds: number }
   if (input.captureDeadline !== undefined) {
+    // Clock-free guard on the absolute opt-in: NaN serializes to null on the
+    // wire and fractional/non-positive values are invalid Unix seconds. We do
+    // NOT reject a past deadline (< now) — that needs the wall clock and is
+    // deliberately deferred to the facilitator.
+    if (
+      !Number.isSafeInteger(input.captureDeadline) ||
+      input.captureDeadline <= 0
+    ) {
+      throw new ValidationError(
+        `captureDeadline (${input.captureDeadline}) must be a positive integer (Unix seconds)`,
+      )
+    }
     capture = { captureDeadline: input.captureDeadline }
   } else {
     const captureDeadlineSeconds =
@@ -195,6 +207,14 @@ export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
 
   let refund: { refundDeadline: number } | { refundDeadlineSeconds: number }
   if (input.refundDeadline !== undefined) {
+    if (
+      !Number.isSafeInteger(input.refundDeadline) ||
+      input.refundDeadline <= 0
+    ) {
+      throw new ValidationError(
+        `refundDeadline (${input.refundDeadline}) must be a positive integer (Unix seconds)`,
+      )
+    }
     refund = { refundDeadline: input.refundDeadline }
   } else {
     const refundDeadlineSeconds =
--- a/packages/helpers/src/x402r-defaults.ts
+++ b/packages/helpers/src/x402r-defaults.ts
@@ -182,9 +182,12 @@ export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
     // A non-positive relative offset resolves to a deadline at or before now,
     // which is never what the caller intends. This is clock-free: we reject the
     // offset itself, not a computed wall-clock deadline.
-    if (captureDeadlineSeconds <= 0) {
+    if (
+      !Number.isSafeInteger(captureDeadlineSeconds) ||
+      captureDeadlineSeconds <= 0
+    ) {
       throw new ValidationError(
-        `captureDeadlineSeconds (${captureDeadlineSeconds}) must be a positive number of seconds: a non-positive offset resolves to a capture deadline at or before now`,
+        `captureDeadlineSeconds (${captureDeadlineSeconds}) must be a positive integer number of seconds: a non-positive offset resolves to a capture deadline at or before now`,
       )
     }
     capture = { captureDeadlineSeconds }
@@ -196,9 +199,12 @@ export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
   } else {
     const refundDeadlineSeconds =
       input.refundDeadlineSeconds ?? DEFAULT_REFUND_WINDOW_SECONDS
-    if (refundDeadlineSeconds <= 0) {
+    if (
+      !Number.isSafeInteger(refundDeadlineSeconds) ||
+      refundDeadlineSeconds <= 0
+    ) {
       throw new ValidationError(
-        `refundDeadlineSeconds (${refundDeadlineSeconds}) must be a positive number of seconds: a non-positive offset resolves to a refund deadline at or before now`,
+        `refundDeadlineSeconds (${refundDeadlineSeconds}) must be a positive integer number of seconds: a non-positive offset resolves to a refund deadline at or before now`,
       )
     }
     refund = { refundDeadlineSeconds }
--- a/packages/helpers/src/x402r-defaults.ts
+++ b/packages/helpers/src/x402r-defaults.ts
@@ -22,7 +22,7 @@ const DEFAULT_TOKEN_VERSION = '2'
  * them to absolute deadlines before the wire payload is built). We model the
  * relative variant here, in helpers, rather than widening the wire type.
  */
-type DeadlineExtra = (
+export type DeadlineExtra = (
   | { captureDeadline: number }
   | { captureDeadlineSeconds: number }
 ) &
@@ -85,12 +85,18 @@ export interface X402rDefaultsInput {
    * which the server scheme refreshes per request. Resolved independently of
    * the refund window — set this to fix the capture deadline while leaving the
    * refund window relative (or vice versa).
+   *
+   * If both this and `captureDeadlineSeconds` are set, the absolute value
+   * wins and the relative offset is ignored.
    */
   captureDeadline?: number
   /**
    * Absolute Unix seconds; refunds allowed until this. Opt-in escape hatch —
    * see `captureDeadline`. Prefer `refundDeadlineSeconds`. Resolved
    * independently of the capture window.
+   *
+   * If both this and `refundDeadlineSeconds` are set, the absolute value wins
+   * and the relative offset is ignored.
    */
   refundDeadline?: number
   /** Floor on the captureAuthorizer's fee in basis points. Defaults to `0` (no minimum). */
@@ -167,20 +173,36 @@ export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
   // value when the caller passed it, else its relative offset — which the
   // server scheme refreshes per request, avoiding the wall-clock-frozen-at-boot
   // footgun the absolute defaults used to cause.
-  const capture =
-    input.captureDeadline !== undefined
-      ? { captureDeadline: input.captureDeadline }
-      : {
-          captureDeadlineSeconds:
-            input.captureDeadlineSeconds ?? DEFAULT_CAPTURE_WINDOW_SECONDS,
-        }
-  const refund =
-    input.refundDeadline !== undefined
-      ? { refundDeadline: input.refundDeadline }
-      : {
-          refundDeadlineSeconds:
-            input.refundDeadlineSeconds ?? DEFAULT_REFUND_WINDOW_SECONDS,
-        }
+  let capture: { captureDeadline: number } | { captureDeadlineSeconds: number }
+  if (input.captureDeadline !== undefined) {
+    capture = { captureDeadline: input.captureDeadline }
+  } else {
+    const captureDeadlineSeconds =
+      input.captureDeadlineSeconds ?? DEFAULT_CAPTURE_WINDOW_SECONDS
+    // A non-positive relative offset resolves to a deadline at or before now,
+    // which is never what the caller intends. This is clock-free: we reject the
+    // offset itself, not a computed wall-clock deadline.
+    if (captureDeadlineSeconds <= 0) {
+      throw new ValidationError(
+        `captureDeadlineSeconds (${captureDeadlineSeconds}) must be a positive number of seconds: a non-positive offset resolves to a capture deadline at or before now`,
+      )
+    }
+    capture = { captureDeadlineSeconds }
+  }
+
+  let refund: { refundDeadline: number } | { refundDeadlineSeconds: number }
+  if (input.refundDeadline !== undefined) {
+    refund = { refundDeadline: input.refundDeadline }
+  } else {
+    const refundDeadlineSeconds =
+      input.refundDeadlineSeconds ?? DEFAULT_REFUND_WINDOW_SECONDS
+    if (refundDeadlineSeconds <= 0) {
+      throw new ValidationError(
+        `refundDeadlineSeconds (${refundDeadlineSeconds}) must be a positive number of seconds: a non-positive offset resolves to a refund deadline at or before now`,
+      )
+    }
+    refund = { refundDeadlineSeconds }
+  }
 
   // Guard ordering only when it's comparable without reading the clock: both
   // absolute, or both relative. The refund window must not end before the
@@ -200,8 +222,6 @@ export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
   if (
     'captureDeadlineSeconds' in capture &&
     'refundDeadlineSeconds' in refund &&
-    capture.captureDeadlineSeconds !== undefined &&
-    refund.refundDeadlineSeconds !== undefined &&
     refund.refundDeadlineSeconds < capture.captureDeadlineSeconds
   ) {
     throw new ValidationError(
--- a/packages/helpers/src/x402r-defaults.ts
+++ b/packages/helpers/src/x402r-defaults.ts
@@ -1,3 +1,4 @@
+import { ValidationError } from '@x402r/core/errors'
 import type { AuthCaptureExtra } from '@x402r/evm'
 
 const DEFAULT_CAPTURE_WINDOW_SECONDS = 60 * 60 * 24 // 24 hours
@@ -7,12 +8,49 @@ const DEFAULT_MAX_FEE_BPS = 100 // 1%
 const DEFAULT_TOKEN_NAME = 'USDC'
 const DEFAULT_TOKEN_VERSION = '2'
 
+/**
+ * The deadline fields `x402rDefaults` may emit. Capture and refund are resolved
+ * independently: each is either its absolute Unix-seconds deadline (opt-in) or
+ * its relative offset (`*Seconds`, the default). The two windows can use
+ * different shapes — e.g. an absolute `captureDeadline` alongside a relative
+ * `refundDeadlineSeconds`.
+ *
+ * `@x402r/evm`'s `AuthCaptureExtra` types only the absolute `captureDeadline` /
+ * `refundDeadline` and has no `*Seconds` fields, because those are a
+ * merchant-input convenience the server scheme resolves away per request (it
+ * reads `extra.captureDeadlineSeconds` / `refundDeadlineSeconds` and converts
+ * them to absolute deadlines before the wire payload is built). We model the
+ * relative variant here, in helpers, rather than widening the wire type.
+ */
+type DeadlineExtra = (
+  | { captureDeadline: number }
+  | { captureDeadlineSeconds: number }
+) &
+  ({ refundDeadline: number } | { refundDeadlineSeconds: number })
+
+/**
+ * Return shape of {@link x402rDefaults}: every non-deadline `AuthCaptureExtra`
+ * field, plus a per-field deadline product — each of capture / refund is either
+ * its absolute deadline or its relative `*Seconds` offset.
+ */
+export type X402rDefaultsExtra = Omit<
+  AuthCaptureExtra,
+  'captureDeadline' | 'refundDeadline'
+> &
+  DeadlineExtra
+
 // Address fields below use the template-literal type `0x${string}` (the
 // exact shape used by AuthCaptureExtra in @x402r/evm@0.2.0-alpha.0).
 // Mirrors upstream rather than importing viem's branded Address, which
 // would pull viem into @x402r/helpers' dep surface for marginal value.
 export interface X402rDefaultsInput {
-  /** Address allowed to call authorize/capture/void/refund/charge on AuthCaptureEscrow. Facilitator-specific. */
+  /**
+   * Address allowed to call authorize/capture/void/refund/charge on
+   * AuthCaptureEscrow. Facilitator-specific. Becomes the on-chain
+   * `PaymentInfo.operator`; it may be a facilitator EOA or a deployed
+   * `PaymentOperator` contract — not necessarily this SDK's deployed
+   * `operatorAddress`.
+   */
   captureAuthorizer: `0x${string}`
   /**
    * Address that receives the fee portion of every settlement. Defaults to
@@ -27,15 +65,33 @@ export interface X402rDefaultsInput {
    */
   feeRecipient?: `0x${string}`
   /**
-   * Absolute Unix seconds; capture must occur before this. Defaults to
-   * `now + 24 hours`.
+   * Relative capture window in seconds. The server scheme resolves this to an
+   * absolute deadline per request, so every authorization gets a fresh window.
+   * Defaults to `86400` (24 hours).
    *
    * Quick-start default favors async capture pipelines (worker queues,
    * manual review, batched capture). Tighten the override for atomic flows
    * where capture happens immediately after authorize.
    */
+  captureDeadlineSeconds?: number
+  /**
+   * Relative refund window in seconds, resolved to an absolute deadline per
+   * request by the server scheme. Defaults to `604800` (7 days).
+   */
+  refundDeadlineSeconds?: number
+  /**
+   * Absolute Unix seconds; capture must occur before this. Opt-in escape
+   * hatch for a fixed wall-clock deadline; prefer `captureDeadlineSeconds`,
+   * which the server scheme refreshes per request. Resolved independently of
+   * the refund window — set this to fix the capture deadline while leaving the
+   * refund window relative (or vice versa).
+   */
   captureDeadline?: number
-  /** Absolute Unix seconds; refunds allowed until this. Defaults to `now + 7 days`. */
+  /**
+   * Absolute Unix seconds; refunds allowed until this. Opt-in escape hatch —
+   * see `captureDeadline`. Prefer `refundDeadlineSeconds`. Resolved
+   * independently of the capture window.
+   */
   refundDeadline?: number
   /** Floor on the captureAuthorizer's fee in basis points. Defaults to `0` (no minimum). */
   minFeeBps?: number
@@ -70,7 +126,7 @@ export interface X402rDefaultsInput {
 }
 
 /**
- * Builds an `AuthCaptureExtra` with x402r's quick-start defaults.
+ * Builds an auth-capture `extra` with x402r's quick-start defaults.
  *
  * **Audience.** Merchants building `PaymentRequirements` directly
  * (single-tenant facilitator, merchant-as-facilitator atomic-charge pattern)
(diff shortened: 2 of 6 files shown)


vraspar:
### SDK Review

Found 2 issues (axes: SDK design, docs/API alignment, tests):

1. **[SDK design]** Partial absolute deadline inputs are silently discarded. `X402rDefaultsInput` allows `captureDeadline` and `refundDeadline` independently, but the implementation only honors absolute deadlines when both are present; a caller who supplies only one explicit absolute deadline gets relative defaults instead.

   https://github.com/BackTrackCo/x402r-sdk/blob/b141688fa0a2b22929da6ea59f4854d51f0863e7/packages/helpers/src/x402r-defaults.ts#L156-L167

2. **[Docs/API alignment]** The README still describes `x402rDefaults` as producing wire-format `AuthCaptureExtra`, but the default output is now server-side pre-wire input with relative deadline offsets. Please update this to say the server scheme resolves `captureDeadlineSeconds` / `refundDeadlineSeconds` into absolute wire deadlines before the payer client signs.

   https://github.com/BackTrackCo/x402r-sdk/blob/b141688fa0a2b22929da6ea59f4854d51f0863e7/packages/helpers/README.md#L24-L32

**Suggestions / nits:**

- The core deadline design looks right: helper emits relative policy windows, `@x402r/evm/auth-capture/server` resolves them per request, and the payer client signs finalized absolute deadlines. The confusing part is documentation/naming, not the approach.
- Consider making the deadline input shape explicit as a mode: relative pair or absolute pair, with no partial/mixed shape. If mixed support is intentional, mirror the server scheme behavior and add tests for absolute capture + relative refund, and relative capture + absolute refund.
- Test coverage is mostly meaningful. The weaker test is the determinism check because two immediate calls can pass even with a `Date.now()` implementation if they happen in the same second; a `Date.now` spy or the output-shape assertions would be stronger.
- Comments/JSDoc are verbose but mostly useful because the server-input vs wire-format distinction is subtle. The test comments can be trimmed first.
- Please clarify `captureAuthorizer` wording: it becomes `PaymentInfo.operator`, but it is not necessarily the SDK's deployed `operatorAddress`. It can be a facilitator EOA in simple flows or a deployed `PaymentOperator` when the merchant wants operator policy/hooks.
- The changeset is accurate but dense. Splitting the sentence and dropping `Closes #183` from published changelog text would read cleaner.

---
Generated with [Codex](https://Codex.ai/code) using review-sdk skill

vraspar:
Thanks for the review. Pushed fixes (amended):

Accepted/fixed:
- Partial absolute deadlines no longer discarded: capture and refund now resolve independently — each field emits its absolute value if you pass it, else its relative `*Seconds` offset. Added tests for absolute-capture/relative-refund and relative-capture/absolute-refund. This is the "mirror the server scheme + add mixed tests" path you suggested (the server resolves each deadline independently, so the helper does too).
- README: dropped the "wire-format AuthCaptureExtra" framing; it now says server-side pre-wire input and explains the server scheme resolves `captureDeadlineSeconds`/`refundDeadlineSeconds` to absolute wire deadlines per request before the payer client signs.
- Determinism test strengthened with a `Date.now` spy asserting it's never called (proves wall-clock-free), kept alongside the deep-equal check.
- `captureAuthorizer` JSDoc clarified: becomes `PaymentInfo.operator`; may be a facilitator EOA or a deployed `PaymentOperator`, not necessarily the SDK's deployed `operatorAddress`.
- Changeset split into two sentences and `Closes #183` dropped from the published text.
- Trimmed the redundant test comments.

Pushback (kept by design):
- I went with per-field (mixed-capable) resolution rather than a strict relative-pair/absolute-pair mode, because it mirrors the auth-capture server scheme, which accepts mixed shapes and resolves each deadline independently. Mixed-shape ordering is intentionally not validated in-helper: absolute vs relative aren't comparable without reading the wall clock (the helper's determinism guarantee), so that ordering check is deferred to the facilitator post-resolution; both-absolute and both-relative are still guarded. There's a test documenting the mixed-skip-validation behavior on purpose.

Tests: 36 pass; typecheck and the build/typecheck pre-commit hook are green.

tofu-miso (commented):
## Overview

Solid fix for a real footgun. The absolute-deadline-at-startup bug was subtle and this change addresses it cleanly. The type design (`DeadlineExtra` as a product of two unions, independent resolution) is well-thought-out. JSDoc and test coverage are thorough.

A few things worth addressing before merge:

**Bug / undocumented behavior:** If a caller passes *both* `captureDeadline` and `captureDeadlineSeconds`, `captureDeadline` silently wins — same for the refund pair. This is currently unspecified in the JSDoc and untested. A user could pass both (e.g. for readability or by mistake) and get surprising output. Either document the precedence explicitly or throw a `ValidationError` like you do for ordering violations.

**Redundant narrowing guards:** In the relative-ordering check, the `!== undefined` conditions after the `in`-narrowing are always true — TypeScript narrows `capture.captureDeadlineSeconds` to `number` once `"captureDeadlineSeconds" in capture` is true (given the discriminated `capture` type). Safe to drop them.

**Missing edge-case test:** `captureDeadlineSeconds: 0` or `refundDeadlineSeconds: 0` — the nullish coalescing (`?? DEFAULT_*`) correctly preserves `0`, but there is no test confirming this, and `0` is a surprising but valid (if unusual) input.

**Nit — test import:** `vi` and `afterEach` are imported at the top level but only used inside the `deadlines` describe block. Not a bug, just noise for readers of other blocks.

**Non-blocking — `DeadlineExtra` not exported:** Consumers who want to extend or narrow `X402rDefaultsExtra` would need to re-derive the deadline union themselves. Consider re-exporting `DeadlineExtra` if downstream packages are expected to use the type directly. Fine to leave as a follow-up.

packages/helpers/src/x402r-defaults.ts
```diff
@@ -97,14 +162,57 @@ export interface X402rDefaultsInput {
  * Optional flags (`autoCapture`, `assetTransferMethod`) are omitted from the
  * output when undefined so the facilitator's wire-spec defaults take over.
  */
-export function x402rDefaults(input: X402rDefaultsInput): AuthCaptureExtra {
-  const nowSeconds = Math.floor(Date.now() / 1000)
+export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
+  // Resolve capture and refund independently. Each field emits its absolute
```
tofu-miso, line 172: **Bug / unspecified precedence:** If a caller passes both `captureDeadline` and `captureDeadlineSeconds`, `captureDeadline` silently wins because the ternary checks `!== undefined` first. Same applies to the refund pair. This should either be documented explicitly in the JSDoc or guarded with a `ValidationError` — silent precedence on conflicting inputs is a footgun identical in spirit to the one this PR is fixing.

packages/helpers/src/x402r-defaults.ts
```diff
@@ -97,14 +162,57 @@ export interface X402rDefaultsInput {
...
+  if (
+    input.captureDeadline !== undefined &&
+    input.refundDeadline !== undefined &&
+    input.refundDeadline < input.captureDeadline
+  ) {
+    throw new ValidationError(
+      `refundDeadline (${input.refundDeadline}) must not be earlier than captureDeadline (${input.captureDeadline}): the refund window cannot end before the capture deadline`,
+    )
+  }
+  if (
```
tofu-miso, line 248: **Nit — redundant guards:** The `capture.captureDeadlineSeconds !== undefined` and `refund.refundDeadlineSeconds !== undefined` checks are unnecessary. Once `"captureDeadlineSeconds" in capture` is true, TypeScript narrows `capture` to `{ captureDeadlineSeconds: number }`, so the value is always a `number`. Same for the refund side. Safe to remove the two `!== undefined` conditions to reduce noise.

packages/helpers/tests/x402r-defaults.test.ts
```diff
@@ -54,15 +57,128 @@ describe('x402rDefaults', () => {
...
 
-    it('produces a valid AuthCaptureExtra with only captureAuthorizer', () => {
-      expect(isAuthCaptureExtra(x402rDefaults(minimalInput))).toBe(true)
+    it('emits relative offsets by default and never reads the wall clock', () => {
+      const nowSpy = vi.spyOn(Date, 'now')
+      const extra = x402rDefaults(minimalInput)
+      expect(extra).toMatchObject({
+        captureDeadlineSeconds: 86_400, // 24 hours
+        refundDeadlineSeconds: 604_800, // 7 days
+      })
```
tofu-miso, line 75: **Nit — missing edge-case test:** `captureDeadlineSeconds: 0` (and `refundDeadlineSeconds: 0`) should be preserved as-is since you use `?? DEFAULT_*` (nullish coalescing, not `||`). A quick test confirming `0` is not replaced by the default would close the loop on this.

packages/helpers/src/x402r-defaults.ts
```diff
@@ -97,14 +162,57 @@ export interface X402rDefaultsInput {
...
+    'captureDeadlineSeconds' in capture &&
+    'refundDeadlineSeconds' in refund &&
+    capture.captureDeadlineSeconds !== undefined &&
+    refund.refundDeadlineSeconds !== undefined &&
```
tofu-miso: Nit: these two `!== undefined` guards are dead. When `'captureDeadlineSeconds' in capture` is true, the field was assigned `input.captureDeadlineSeconds ?? DEFAULT_CAPTURE_WINDOW_SECONDS` (lines 174-175) — always a defined number; likewise for refund. The `in` checks alone are sufficient, so the two `!== undefined` lines can be dropped.

packages/helpers/src/x402r-defaults.ts
```diff
@@ -97,14 +162,57 @@ export interface X402rDefaultsInput {
...
  */
-export function x402rDefaults(input: X402rDefaultsInput): AuthCaptureExtra {
-  const nowSeconds = Math.floor(Date.now() / 1000)
+export function x402rDefaults(input: X402rDefaultsInput): X402rDefaultsExtra {
+  // Resolve capture and refund independently. Each field emits its absolute
+  // value when the caller passed it, else its relative offset — which the
+  // server scheme refreshes per request, avoiding the wall-clock-frozen-at-boot
+  // footgun the absolute defaults used to cause.
+  const capture =
+    input.captureDeadline !== undefined
```
tofu-miso: Minor: `X402rDefaultsInput` lets a caller pass both `captureDeadline` (absolute) and `captureDeadlineSeconds` (relative) for the same window. Here the absolute branch wins and the relative value is silently ignored — no warning, no type error (same for refund). Consider documenting the precedence on the field JSDoc, or narrowing the input type so the two are mutually exclusive, so a caller who sets both isn't surprised.

packages/helpers/src/x402r-defaults.ts
```diff
@@ -97,14 +162,57 @@ export interface X402rDefaultsInput {
...
+      ? { captureDeadline: input.captureDeadline }
+      : {
+          captureDeadlineSeconds:
+            input.captureDeadlineSeconds ?? DEFAULT_CAPTURE_WINDOW_SECONDS,
```
tofu-miso: Minor: there's no lower-bound validation on the offsets. A negative or zero `captureDeadlineSeconds` / `refundDeadlineSeconds` resolves to a deadline at/before `now` once the server adds it, and a past absolute `captureDeadline` is likewise accepted. The ordering guard catches refund < capture but not a non-positive window. If the facilitator already rejects these downstream this is fine to leave, but a cheap `> 0` check here would fail closer to the mistake.

packages/helpers/src/x402r-defaults.ts
```diff
@@ -7,12 +8,49 @@ const DEFAULT_MAX_FEE_BPS = 100 // 1%
...
+ * `refundDeadlineSeconds`.
+ *
+ * `@x402r/evm`'s `AuthCaptureExtra` types only the absolute `captureDeadline` /
+ * `refundDeadline` and has no `*Seconds` fields, because those are a
+ * merchant-input convenience the server scheme resolves away per request (it
+ * reads `extra.captureDeadlineSeconds` / `refundDeadlineSeconds` and converts
+ * them to absolute deadlines before the wire payload is built). We model the
+ * relative variant here, in helpers, rather than widening the wire type.
+ */
+type DeadlineExtra = (
```
tofu-miso: Nit (informational): this union is structurally non-exclusive — an object carrying both `captureDeadline` and `captureDeadlineSeconds` still satisfies `{ captureDeadline: number } | { captureDeadlineSeconds: number }`, so the type doesn't actually forbid emitting both. The runtime construction guarantees exactly one, so it's fine in practice; just noting the type isn't load-bearing for that invariant.

.changeset/helpers-x402rdefaults-relative-offsets.md
```diff
@@ -0,0 +1,5 @@
+---
+"@x402r/helpers": minor
```
tofu-miso, line 2: The default return shape changes: callers reading `.captureDeadline` / `.refundDeadline` off `x402rDefaults(...)` now get `undefined` by default, and the return type changed from `AuthCaptureExtra` to `X402rDefaultsExtra`. That's behaviorally breaking for existing consumers. Under 0.x-alpha semver a `minor` is defensible, but worth a conscious confirmation that's the intended classification given the output contract changed.

packages/helpers/tests/x402r-defaults.test.ts
```diff
@@ -54,15 +57,128 @@ describe('x402rDefaults', () => {
...
+      expect('captureDeadline' in extra).toBe(false)
+      expect('refundDeadline' in extra).toBe(false)
+      // Proves the helper is wall-clock-free, so its output is deterministic.
+      expect(nowSpy).not.toHaveBeenCalled()
     })
 
+    it('is deterministic — same input yields byte-identical output', () => {
```
tofu-miso: Nit: `toEqual` is a structural deep-equality check, not 'byte-identical' (nothing is serialized/compared byte-wise). The determinism property being asserted is real; the title just slightly overstates it — 'deep-equal output' would be more precise.

tofu-miso (commented):
Review of the relative-deadline-offsets change. The core fix (relative-by-default, wall-clock-free, per-field absolute opt-in) is well-designed, the JSDoc/README are thorough, and the test matrix covers the new behavior well. Findings below are mostly nits and minor discussion items — no blocking correctness bugs found.

vraspar:
Thanks @tofu-miso — addressed in a follow-up commit (`2b031b4`, no force-push):

- **Precedence (#1):** documented on `captureDeadline`/`refundDeadline` that the absolute value wins over its `*Seconds` counterpart when both are set.
- **Dead guards (#2):** removed the two redundant `!== undefined` checks (already narrowed by the `in` checks).
- **Positive-offset guard (#6, supersedes #3):** `captureDeadlineSeconds`/`refundDeadlineSeconds` ≤ 0 now throws `ValidationError` (clock-free — a non-positive offset resolves to a deadline at/before now). I went with *rejecting* a non-positive offset rather than the "preserve `0`" test, since `0` is a footgun, not a meaningful value. Tested for `0` and negative.
- **Export `DeadlineExtra` (#5):** now exported so consumers can compose `X402rDefaultsExtra`.
- **Test title (#9):** "byte-identical" → "deep-equal output" (the assertion is `toEqual`).

Not changing (with reasons):

- **Top-level `vi`/`afterEach` imports (#4):** that's the only valid location for an `import` — it can't be scoped to a `describe` block.
- **`DeadlineExtra` structural non-exclusivity (#7):** a known TypeScript intersection-of-unions limitation; the runtime construction emits exactly one shape per field, so there's no defect to fix (flagged as informational).
- **`minor` vs `major` (#8):** `minor` is the repo's 0.x-alpha convention for breaking changes; `major` would be incorrect, and I don't bump versions without an explicit instruction.

Tests: 40 pass (+4 guard tests); typecheck + build green.

packages/helpers/src/x402r-defaults.ts
```diff
@@ -97,14 +168,71 @@ export interface X402rDefaultsInput {
...
+  let capture: { captureDeadline: number } | { captureDeadlineSeconds: number }
+  if (input.captureDeadline !== undefined) {
+    capture = { captureDeadline: input.captureDeadline }
+  } else {
+    const captureDeadlineSeconds =
+      input.captureDeadlineSeconds ?? DEFAULT_CAPTURE_WINDOW_SECONDS
+    // A non-positive relative offset resolves to a deadline at or before now,
+    // which is never what the caller intends. This is clock-free: we reject the
+    // offset itself, not a computed wall-clock deadline.
+    if (captureDeadlineSeconds <= 0) {
```
tofu-miso: Nit — the `<= 0` guard doesn't fully enforce its stated contract ("must be a positive number of seconds"). `NaN` slips through: `NaN <= 0` is `false`, so `x402rDefaults({ captureAuthorizer, captureDeadlineSeconds: NaN })` emits `{ captureDeadlineSeconds: NaN }`, which serializes to `null` on the wire — a silent corruption of exactly the kind this guard exists to prevent. `Number(process.env.CAPTURE_SECONDS)` for an unset/typo'd env var is the realistic path in. Fractional offsets (`0.5`) also pass, though Unix-second offsets should be integers. (`Infinity` only throws incidentally, via the ordering check against the 604800 refund default — set both to `Infinity` and it'd pass.)

A `!Number.isInteger(captureDeadlineSeconds) || captureDeadlineSeconds <= 0` (or `Number.isSafeInteger`) check would close it and is still clock-free. Same for the refund guard below at L199.

packages/helpers/tests/x402r-defaults.test.ts
```diff
@@ -54,15 +57,153 @@ describe('x402rDefaults', () => {
...
+      })
+      expect(extra).toMatchObject({
+        captureDeadlineSeconds: 30 * 86_400,
+        refundDeadlineSeconds: 60 * 86_400,
+      })
+      expect('captureDeadline' in extra).toBe(false)
+      expect('refundDeadline' in extra).toBe(false)
+    })
+
+    it('emits both absolute when both are supplied', () => {
```
tofu-miso, line 100: Nit — the precedence you documented for the both-set case (absolute wins over `*Seconds`) is now spelled out in the JSDoc but still has no test locking it in. The original review flagged it as "unspecified in the JSDoc **and** untested"; the doc half is done, the test half isn't. A sibling to this test — pass both `captureDeadline` + `captureDeadlineSeconds` (and the refund pair) and assert the absolute is emitted while the `*Seconds` field is absent — would guard against a future refactor silently flipping the precedence.

tofu-miso (commented):
Re-review of the follow-up commit (`2b031b4`). The prior round's nits are all addressed correctly: precedence is now documented on both absolute fields, the two dead `!== undefined` guards are gone, `DeadlineExtra` is exported, the non-positive-offset `ValidationError` guard is in (clock-free, as argued), and the determinism test title now matches the `toEqual` assertion. Tests (40) and typecheck are green locally.

Two small new items below; both are nits, nothing blocking. Your pushback on the kept items (mixed-shape ordering deferred to the facilitator, top-level `vi`/`afterEach` import, structural non-exclusivity of the union, `minor` under 0.x-alpha) all hold up — see the summary in the thread.

vraspar:
Good catch @tofu-miso — addressed in `0225335` (new commit, no force-push):

- **Guards now reject NaN and fractional offsets**, not just `≤ 0`: `!Number.isSafeInteger(x) || x <= 0` for both the capture and refund relative-offset paths (deadlines are integer seconds). This closes the `Number(process.env.X)` → `NaN` → `null`-on-the-wire path. Messages updated to "must be a positive integer number of seconds."
- **Tests added:** `NaN` and fractional (`0.5`) offsets throw `ValidationError` (capture + refund), plus two precedence tests asserting an absolute `captureDeadline`/`refundDeadline` wins over its `*Seconds` counterpart (the documented behavior, now locked).

46 tests pass. No changeset — this only tightens validation of already-invalid input within the feature this PR's changeset already covers.

packages/helpers/src/x402r-defaults.ts
tofu-miso: Verified the guard in `0225335` — `!Number.isSafeInteger(x) || x <= 0` is correct and the NaN/fractional tests lock it in. One asymmetry worth a follow-up though: this guard only sits on the **relative** `*Seconds` branches. The **absolute** opt-in fields (`captureDeadline` / `refundDeadline`) are still passed through unvalidated, so the same `Number(process.env.X)` footgun you just closed is still reachable through the absolute path:

```ts
x402rDefaults({ captureAuthorizer, captureDeadline: NaN })
// → { captureDeadline: null, ... }   // NaN → null on the wire — the exact silent corruption

x402rDefaults({ captureAuthorizer, refundDeadline: 0.5 })   // → { refundDeadline: 0.5 }  (uint48 wants an integer)
x402rDefaults({ captureAuthorizer, captureDeadline: -5 })   // → { captureDeadline: -5 }
```

(I confirmed all three locally against this commit.)

Fair caveat — and where my original comment over-reached: rejecting a *past* absolute deadline does need the wall clock, so that part genuinely can't be done clock-free and should stay deferred to the facilitator. But NaN / non-integer / non-positive on the absolute fields are all clock-free, and a `Number.isSafeInteger(x) && x > 0` check there would close the `NaN → null` vector symmetrically with what you just did for the relative path. Non-blocking, but it's the identical corruption class, so it'd be odd to guard one entry point and not the other.

tofu-miso:
## Re-review — assessment of the pushback (round 4)

Re-checked the branch at `0225335`: built `@x402r/core` + `@x402r/evm`, ran the helpers suite (**46/46 pass**) and `tsc --noEmit` (**green**). The fixes land as described — NaN/fractional guards on the relative offsets, precedence tests that assert the `*Seconds` field is *absent* when the absolute wins, `DeadlineExtra` exported, determinism test with the `Date.now` spy.

**Where the author pushed back, and my call:**

| Item | Author's position | My assessment |
|---|---|---|
| Mixed-shape ordering deferred to facilitator | Can't compare absolute vs relative without the wall clock; would break the determinism guarantee | **Author right.** Both-absolute and both-relative are still guarded; the mixed case genuinely isn't comparable clock-free. Correct to defer. |
| Top-level `vi`/`afterEach` import | A static `import` can't be scoped to a `describe` block | **Author right.** ES static imports are module-level only; the only alternative is dynamic `import()`, which is worse for test utilities. My nit isn't actionable as written. |
| `DeadlineExtra` structural non-exclusivity | TS intersection-of-unions can't forbid both keys; runtime construction guarantees one | **Author right** — and it was informational to begin with. |
| `minor` vs `major` under 0.x-alpha | 0.x convention; breaking changes ride a minor | **Author right** (semver §4). My own comment already called the minor "defensible." |
| Reject `0`/non-positive offset instead of "preserve 0" | `0` is a footgun (resolves to a deadline at now), not a meaningful value | **Author right, my original nit was the weaker call.** Testing that `0` is *preserved* would have locked in bad behavior; rejecting it fails closer to the mistake. Good supersession. |

So: **no incorrect pushback.** Every kept-by-design item holds up, and the accepted items were resolved correctly and verified.

**One residual gap** (posted inline on the NaN-guard thread, [r3369930534](https://github.com/BackTrackCo/x402r-sdk/pull/190#discussion_r3369930534)): the `Number.isSafeInteger / > 0` guard `0225335` added covers only the **relative** `*Seconds` branches. The **absolute** `captureDeadline` / `refundDeadline` opt-in fields are still unvalidated, so `captureDeadline: Number(process.env.X)` → `NaN` → `null` on the wire — the exact corruption that commit set out to close, reachable through the other entry point. Confirmed locally. Clock-free to fix (NaN/integer/positivity only; "past deadline" rightly stays deferred). Non-blocking follow-up.

Otherwise this is good to merge.

A1igator:
### SDK Review: Round 2

Found 2 new issues since commit b141688 (axes: tests, conventions, dead code, example alignment, SDK design):

**Resolved since last round:** partial absolute deadline inputs silently discarded (now resolved independently per field, with mixed-shape tests); README describing the output as wire-format `AuthCaptureExtra`.
**Still open:** none.
**New:**

1. **[Stale docs]** The pending changeset from PR #123 still reads `x402rDefaults(input) → AuthCaptureExtra`. The repo is in changesets pre-release mode, so this file re-renders into the next stable release notes, and this PR makes that signature wrong: readers will expect `captureDeadline` on the default output, which no longer carries it. One-line fix to `X402rDefaultsExtra`.

   https://github.com/BackTrackCo/x402r-sdk/blob/022533584ea6027180d8d85bef7a4bb9349a21a6/.changeset/sdk-authcapture-autocapture-wireformat.md#L17

2. **[Tests]** `DeadlineExtra` and `X402rDefaultsExtra` are exported from the barrel but nothing in-repo imports them from `@x402r/helpers`; tests import from the relative source path. A barrel refactor that drops the re-exports would still pass the suite. One type-import smoke test pins them.

   https://github.com/BackTrackCo/x402r-sdk/blob/022533584ea6027180d8d85bef7a4bb9349a21a6/packages/helpers/src/index.ts#L34-L38

---

vraspar:
Thanks for the thorough re-review @tofu-miso @A1igator — addressed in `74a6184` (new commit, no force-push):

- **Absolute-field guard (B):** the NaN / non-integer / non-positive guard now also covers the **absolute** `captureDeadline` / `refundDeadline` opt-ins, not just the relative `*Seconds` branches — closing the `captureDeadline: NaN` → `null`-on-the-wire path through the other entry point. No past-deadline (`< now`) check (that needs the wall clock, so it stays deferred to the facilitator). +4 tests mirroring the relative-branch ones (50 total pass).
- **Stale changeset (C1):** corrected `sdk-authcapture-autocapture-wireformat.md` — the `x402rDefaults(input) → AuthCaptureExtra` signature is now `→ X402rDefaultsExtra`, since this PR changed the return type and that pending changelog line was re-rendering wrong.

Leaving as-is (with reason):
- **Barrel-import smoke test (C2):** type-only exports can't silently regress — a dropped `export type` is a compile error at any consumer — so an in-repo type smoke test is low value here. Happy to add it if you feel strongly.

A1igator:
### SDK Review: Round 3

Found 0 new issues since commit 0225335 (axes: tests, conventions, dead code):

**Resolved since last round:** pending changeset now reads `x402rDefaults(input) → X402rDefaultsExtra`.
**Still open:** none. The round-2 barrel-import smoke test is closed as declined; the rationale holds, since a dropped `export type` fails loudly at the consumer's compile step rather than corrupting anything silently.
**New:** none. Verified in `74a6184`: the absolute-field guards mirror the relative ones (both arms reachable, no valid documented input rejected, and an invalid absolute with a valid relative sibling throws rather than silently falling back, matching the documented precedence); the 4 new tests assert `ValidationError` correctly.

---

A1igator (approved):
Approving after review-sdk round 3: 0 new issues at 74a6184, all prior rounds resolved or closed with accepted rationale.

PR #197, opened 2026-06-05
ci: pin foundry-toolchain to v1.7.1 (#159)

### LINKED ISSUE #159: Pin foundry-toolchain to specific Foundry version
## Problem

`ci.yml`'s `test-fork` and `abi-drift` jobs use `foundry-rs/foundry-toolchain@<sha>` (SHA-pinned), but the action itself defaults to `version: stable` — meaning it installs the latest stable Foundry release at workflow run time.

A breaking Foundry stable release could cause spurious failures:
- `abi-drift` could fail because Foundry's ABI output changed slightly
- `test-fork` could fail because Anvil behavior changed
- Neither would correspond to any change in our code or contracts

## Proposed fix

Pin `with: version:` on both `foundry-toolchain` step invocations:

```yaml
- name: Install Foundry
  uses: foundry-rs/foundry-toolchain@<sha>
  with:
    version: stable  # or a specific tag like nightly-<commit> or v1.4.0
```

Setting `version: stable` doesn't actually pin (defaults to it anyway). For real pinning, use a specific Foundry release tag.

Need to pick a known-good version. Suggestion: run a recent main CI green, note the Foundry version it installed (`forge --version` output), pin to that.

## Why this matters

- Closes "abi-drift fails for reasons unrelated to ABI drift" failure mode.
- Same defense pattern as SHA-pinning actions: pin the *thing that gets installed*, not just the installer.

## Why this isn't in PR #142

Phase 3 of the release-pipeline rollout (#142) is scoped to **security hardening** (permissions, SHA pins, harden-runner). This is **operational stability** for the build toolchain — different concern, deserves its own PR.

## Related

- Surfaced in #142 audit comment.

vraspar:
## What
Pins the Foundry version installed by `foundry-rs/foundry-toolchain` to `v1.7.1` on both invocations in `ci.yml` (the `test-fork` and `abi-drift` jobs).

## Why
The action is SHA-pinned, but without a `version:` it installs whatever `stable` resolves to at workflow-run time. A breaking Foundry stable release could fail `abi-drift` (ABI output changes) or `test-fork` (Anvil behavior changes) for reasons unrelated to our code. Pinning the installed toolchain — not just the installer — closes that failure mode.

## Version choice
`v1.7.1` is the current latest stable Foundry release and matches what recent green `main` CI installs via `stable`, so this freezes the toolchain at its current state with no behavioral change today. Future bumps are deliberate.

Closes #159

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,8 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@c7450ba673e133f5ee30098b3b54f444d3a2ca2d  # v1.8.0
+        with:
+          version: v1.7.1
       - name: Clone x402r-contracts (with submodules)
         run: git clone --depth 1 --recurse-submodules https://github.com/BackTrackCo/x402r-contracts.git ../x402r-contracts
       - name: Build contracts
@@ -208,6 +210,8 @@ jobs:
         run: git clone --depth 1 --recurse-submodules https://github.com/BackTrackCo/x402r-contracts.git ../x402r-contracts
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@c7450ba673e133f5ee30098b3b54f444d3a2ca2d  # v1.8.0
+        with:
+          version: v1.7.1
       - name: Build contracts
         run: cd ../x402r-contracts && forge build
       - uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093  # v6.0.8
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,11 @@ on:
     branches: [main]
   pull_request:
 
+env:
+  # Single source of truth for the Foundry toolchain version installed by
+  # foundry-rs/foundry-toolchain in the test-fork and abi-drift jobs.
+  FOUNDRY_VERSION: v1.7.1
+
 # Default-deny: per-job permissions opt in to only what they need.
 permissions: {}
 
@@ -118,7 +123,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@c7450ba673e133f5ee30098b3b54f444d3a2ca2d  # v1.8.0
         with:
-          version: v1.7.1
+          version: ${{ env.FOUNDRY_VERSION }}
       - name: Clone x402r-contracts (with submodules)
         run: git clone --depth 1 --recurse-submodules https://github.com/BackTrackCo/x402r-contracts.git ../x402r-contracts
       - name: Build contracts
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -117,6 +117,8 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@c7450ba673e133f5ee30098b3b54f444d3a2ca2d  # v1.8.0
+        with:
+          version: v1.7.1
```
vraspar: nit: Since this same Foundry pin now appears in both CI jobs, consider defining it once, for example a workflow/job env value like `FOUNDRY_VERSION: v1.7.1`, and referencing that from both `foundry-toolchain` steps. Not blocking; it just prevents the two pins from drifting apart later.

vraspar (commented):
One small non-blocking nit.

A1igator:
I rather we notice a failure if stable causes it rather than pin.

vraspar:
@A1igator fair instinct — floating `stable` to catch upstream regressions is what contract libraries do, where forge/anvil is the system under test (forge-std even runs `nightly` in its PR matrix). But here Anvil/forge is a test *fixture* for the TS SDK, not the thing under test, so an upstream change to gas accounting / traces / fork behavior / ABI output reds our CI as noise on PRs that changed nothing.

The closest analog to our setup — viem — pins the toolchain to an exact version (currently `v1.7.1`) for its anvil-backed suite, and Uniswap's v4-core pins even on its merge-to-main branch. So I'd lean pin here.

If you still want the early-warning signal, I can add `stable` as a non-gating (`continue-on-error`) matrix axis so a Foundry regression shows up yellow without blocking PRs — that's forge-std's nightly-in-matrix pattern adapted. Happy to go that route if you'd prefer it over a hard pin.

(I also DRY'd the version into a single `FOUNDRY_VERSION` env per your other note.)

A1igator:
Yeah go with the yellow route

vraspar:
@A1igator after digging into this, I'd actually skip the canary and just ship the pin — two reasons:

1. **The projects closest to our setup just pin, no canary.** viem and Uniswap's v4-core both pin a single Foundry version with no `stable`/`nightly` leg. The nightly-in-matrix pattern is really specific to contract libraries like forge-std whose actual remit is surfacing Foundry regressions — we're a harness consumer of Anvil, not that.

2. **It's more work + cycles than it's worth here.** A non-gating `stable` leg means doubling the abi-drift job, and it'd inherit the same foundryup-403 flakiness we're already hitting, so the canary signal would mostly be noise. An upstream Foundry regression is really forge-std's tripwire to catch, not ours.

So I'd merge the pin (+ the `FOUNDRY_VERSION` DRY) as-is. If you still feel strongly, happy to track the canary as its own issue rather than grow this PR — but I don't think it earns its keep. lmk.

A1igator (approved):

PR #199, opened 2026-06-05
feat(core): add formatError(err) utility (#118)

### LINKED ISSUE #118: @x402r/core: formatViemError(err) utility
## Problem

Every SDK consumer writes the same error-flattening line when surfacing viem/contract errors to users or HTTP responses:

\`\`\`ts
err.shortMessage ?? err.message ?? String(err)
\`\`\`

Three copies in one file in BackTrackCo/arbiter-examples#12 alone: \`arbiter.ts:130, 192, 212\`. Every arbiter, merchant handler, and CLI ends up rewriting it.

## Proposal

Export a \`formatViemError(err: unknown): string\` utility from \`@x402r/core\` (where \`X402rError\` lives — \`packages/core/src/errors/base.ts\`).

v1 can be the one-liner. Centralizing it gives a single place to enrich later:

- unwrap \`ContractFunctionExecutionError\` → \`shortMessage\`
- fall back to \`details\` before \`String(err)\`
- handle \`X402rError\` instances specially (they already have structured messages)

## Acceptance

- \`@x402r/core\` exports \`formatViemError\`.
- arbiter-examples adopts it — replaces the three inline copies in \`arbiter.ts\`.

## Context

Discovered while reviewing BackTrackCo/arbiter-examples#12.

vraspar:
## What
Adds `formatError(err: unknown): string` to `@x402r/core`, exported from the package root and the `./errors` subpath. Flattens any thrown value to a human-readable string in priority order: `X402rError` → its structured `.message`; viem-shaped errors → `shortMessage`; `Error` → `.message`; otherwise `String(err)`.

Centralizes the `err.shortMessage ?? err.message ?? String(err)` one-liner that consumers (arbiters, merchant handlers, CLI) re-implement, and gives one place to enrich later (e.g. unwrapping `ContractFunctionExecutionError`, falling back to `details`).

> Note: issue #118 proposed the name `formatViemError`; this ships as `formatError` since it also handles `X402rError` / `Error` / string / `unknown` — viem's `shortMessage` is just one of four branches.

## Tests
7 vitest cases covering each branch plus edge inputs (`null` / `undefined` / plain object). The `X402rError`-before-`shortMessage` ordering is guarded with a fixture whose `.message` ≠ `.shortMessage`.

## Out of scope
Adopting this in the arbiter examples is a follow-up in that repo.

Closes #118

### DIFF
--- a/.changeset/core-format-viem-error.md
+++ b/.changeset/core-format-viem-error.md
@@ -1,0 +1,5 @@
+---
+"@x402r/core": minor
+---
+
+Add `formatViemError(err)` for flattening viem/contract errors to a string.
--- a/.changeset/core-format-viem-error.md
+++ b/.changeset/core-format-viem-error.md
@@ -1,5 +1,6 @@
 ---
 "@x402r/core": minor
+"@x402r/sdk": minor
 ---
 
 Add `formatError(err)` for flattening any error (viem/contract, `X402rError`, or unknown) to a human-readable string.
--- a/.changeset/core-format-viem-error.md
+++ b/.changeset/core-format-viem-error.md
@@ -2,4 +2,4 @@
 "@x402r/core": minor
 ---
 
-Add `formatViemError(err)` for flattening viem/contract errors to a string.
+Add `formatError(err)` for flattening any error (viem/contract, `X402rError`, or unknown) to a human-readable string.
--- a/packages/core/src/errors/format.ts
+++ b/packages/core/src/errors/format.ts
@@ -1,0 +1,21 @@
+import { X402rError } from './base.js'
+
+/**
+ * Flatten any thrown value to a human-readable string.
+ *
+ * @example
+ * try { await capture(client, args) } catch (err) { res.status(500).json({ error: formatViemError(err) }) }
+ */
+export function formatViemError(err: unknown): string {
+  if (err instanceof X402rError) return err.message
+  if (
+    err !== null &&
+    typeof err === 'object' &&
+    'shortMessage' in err &&
+    typeof (err as { shortMessage: unknown }).shortMessage === 'string'
+  ) {
+    return (err as { shortMessage: string }).shortMessage
+  }
+  if (err instanceof Error) return err.message
+  return String(err)
+}
--- a/packages/core/src/errors/format.ts
+++ b/packages/core/src/errors/format.ts
@@ -4,9 +4,9 @@ import { X402rError } from './base.js'
  * Flatten any thrown value to a human-readable string.
  *
  * @example
- * try { await capture(client, args) } catch (err) { res.status(500).json({ error: formatViemError(err) }) }
+ * try { await capture(client, args) } catch (err) { console.error(formatError(err)) }
  */
-export function formatViemError(err: unknown): string {
+export function formatError(err: unknown): string {
   if (err instanceof X402rError) return err.message
   if (
     err !== null &&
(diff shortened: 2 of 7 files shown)


tofu-miso (commented):
**Code Review Notes**

**Nit — empty `shortMessage` edge case**
If a viem-shaped object has `shortMessage: ''`, the type guard passes (`typeof '' === 'string'`) and the function returns `''` instead of falling through to `.message` / `String(err)`. Not a regression from the `??` idiom this replaces, but for extra robustness consider checking `.trim() !== ''` as well.

**Question — `X402rError.message` verbosity in API responses**
The `@example` in the JSDoc shows this being used in an HTTP 500 handler: `res.status(500).json({ error: formatViemError(err) })`. For an `X402rError`, `.message` includes `metaMessages`, a docs URL, `Details:`, and a version footer (`Version: @x402r/core@x.y.z`). That's fairly verbose/leaky for a client-facing response body. Worth clarifying intent before the follow-up PR adopts this widely — returning `.shortMessage` for the `X402rError` branch might be preferable if the target is API responses. Please document which consumers should expect either way.

**Nit — function name scope**
`formatViemError` actually handles `X402rError`s, plain `Error`s, strings, and `null`/`undefined` — viem is just one of four branches. `formatError` would describe the function better. Cheapest to rename now before downstream adoption spreads.

**Minor — test coverage gaps**
Two small gaps worth filling if you want exhaustive coverage:
1. No explicit test case for a `ContractCallError` or other `X402rError` subclass (it's covered transitively via `instanceof X402rError` but an explicit case would guard against future refactors).
2. No test for the empty-string `shortMessage` edge noted above.

---

Overall: **approve / merge-ready**. The branch ordering decision (viem structural check before `instanceof Error`) is correct and well-tested. The items above are all nits/questions — nothing blocking.

A1igator:
### SDK Review

Found 2 issues (axes: tests, conventions, dead code, example alignment):

1. **[Example alignment]** `formatError` is missing from the selective `@x402r/core` re-export block in the `@x402r/sdk` barrel. That block already re-exports the error classes (`X402rError`, `ContractCallError`, `ConfigError`, `ValidationError`) and peer formatting utilities (`formatFeeBreakdown`), so `@x402r/sdk` consumers get the errors but not their formatter without adding a direct `@x402r/core` dependency.

   https://github.com/BackTrackCo/x402r-sdk/blob/95ac1445068367e20f53d3d9b681918314cf27a4/packages/sdk/src/index.ts#L36-L56

2. **[Tests]** The `shortMessage` branch is only tested with a plain object, never with an `Error` instance that also carries `shortMessage`, which is the actual shape of every viem `BaseError` (extends `Error` and sets `this.shortMessage`, viem `src/errors/base.ts:74`). A plain-object fixture can never reach the `instanceof Error` branch, so reordering the structural check after `instanceof Error` (which would return the verbose `.message` for all real viem errors) still passes the suite. One fixture like `Object.assign(new Error('long message'), { shortMessage: 'short' })` pins the precedence.

   https://github.com/BackTrackCo/x402r-sdk/blob/95ac1445068367e20f53d3d9b681918314cf27a4/packages/core/tests/format-error.test.ts#L11-L14

---

A1igator:
### SDK Review — Round 2

Found 0 new/unresolved issues since commit 95ac144 (axes: tests, conventions, dead code, example alignment):

1. `formatError` missing from the `@x402r/sdk` barrel — fixed in 89acdb5. Verified: `packages/sdk/src/index.ts` line 41, inside the values block sourced from `@x402r/core`; the export chain resolves through core's `exports` map for both the root and `./errors` subpaths.
2. No test for an `Error` instance carrying `shortMessage` (viem `BaseError` shape) — fixed in 89acdb5. The new fixture pins the branch order: moving the structural `shortMessage` check after `instanceof Error` fails the suite.

**Still open:** none.
**New:** none.

---

A1igator (approved):
Round 2 review-sdk pass found no new issues; both round-1 findings verified fixed in 89acdb5.

PR #201, opened 2026-06-07
chore(helpers): drop unused @x402/evm + viem peer dependencies

vraspar:
## What

`@x402r/helpers` declared `@x402/evm` and `viem` as `peerDependencies`, but its source imports only `@x402r/evm` (the `r` one) and `@x402/core/server` — `viem`/`@x402/evm` were redundant. This drops them from both `peerDependencies` and `devDependencies`.

Consumers' required-peer surface is **unchanged**: `@x402r/helpers` takes `@x402r/evm` as a peer, and `@x402r/evm@0.2.0-alpha.1` already declares `@x402/evm` + `viem` as *its* peers — so a consumer still surfaces them, just via `@x402r/evm` rather than a redundant re-declaration here. No internal workspace package depends on `@x402r/helpers`.

Follow-up to the peer-hygiene note on #185.

## Validation (Node 22 / engine-strict)

- `pnpm --filter @x402r/helpers test` → 30/30 pass
- `typecheck` + `build` (publint + attw) clean — generated `dist/index.d.ts` carries no `viem` import, so nothing leaks into the published types
- `pnpm knip` → exit 0 (no unused/unlisted-dependency error)
- `pnpm install --frozen-lockfile` settled; lockfile diff is surgical (only the helpers importer's `@x402/evm` + `viem` entries removed)

`patch` changeset for `@x402r/helpers`.

### DIFF
--- a/.changeset/helpers-drop-unused-peers.md
+++ b/.changeset/helpers-drop-unused-peers.md
@@ -1,0 +1,5 @@
+---
+"@x402r/helpers": patch
+---
+
+Drop the unused `@x402/evm` and `viem` peer dependencies — `@x402r/helpers` only requires `@x402/core` and `@x402r/evm` (which pulls the rest transitively).
--- a/.changeset/helpers-drop-unused-peers.md
+++ b/.changeset/helpers-drop-unused-peers.md
@@ -2,4 +2,4 @@
 "@x402r/helpers": patch
 ---
 
-Drop the unused `@x402/evm` and `viem` peer dependencies — `@x402r/helpers` only requires `@x402/core` and `@x402r/evm` (which pulls the rest transitively).
+Drop the unused `@x402/evm` and `viem` peer dependencies — `@x402r/helpers` only requires `@x402/core` and `@x402r/evm`. Consumers still provide `viem` / `@x402/evm` via `@x402r/evm`'s own peer requirements.
--- a/packages/helpers/package.json
+++ b/packages/helpers/package.json
@@ -35,15 +35,11 @@
   },
   "peerDependencies": {
     "@x402/core": "^2.14.0",
-    "@x402/evm": "^2.14.0",
-    "@x402r/evm": ">=0.2.0-alpha.1 <0.3.0",
-    "viem": "^2.48.11"
+    "@x402r/evm": ">=0.2.0-alpha.1 <0.3.0"
   },
   "devDependencies": {
     "@x402/core": "2.14.0",
-    "@x402/evm": "2.14.0",
-    "@x402r/evm": "0.2.0-alpha.1",
-    "viem": "^2.48.11"
+    "@x402r/evm": "0.2.0-alpha.1"
   },
   "size-limit": [
     {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -140,15 +140,9 @@ importers:
       '@x402/core':
         specifier: 2.14.0
         version: 2.14.0
-      '@x402/evm':
-        specifier: 2.14.0
-        version: 2.14.0(typescript@6.0.3)
       '@x402r/evm':
         specifier: 0.2.0-alpha.1
         version: 0.2.0-alpha.1(@x402/core@2.14.0)(@x402/evm@2.14.0(typescript@6.0.3))(viem@2.52.0(typescript@6.0.3)(zod@4.4.3))
-      viem:
-        specifier: ^2.48.11
-        version: 2.52.0(typescript@6.0.3)(zod@4.4.3)
 
   packages/sdk:
     dependencies:


tofu-miso (approved):
## Review: chore(helpers): drop unused @x402/evm + viem peer dependencies

**LGTM overall — clean hygiene patch.**

### What this does

Removes `@x402/evm` and `viem` from `@x402r/helpers`' `peerDependencies` and `devDependencies`. These were never imported directly by `helpers` — only `@x402r/evm` and `@x402/core` are. Consumers still surface `viem` / `@x402/evm` via `@x402r/evm`'s own peer chain.

---

### Observations

**`peerDependencies` — correct after change**
The remaining two peers (`@x402/core` and `@x402r/evm`) match what the source actually imports. No phantom deps remaining.

**lockfile — correctly surgical**
The `@x402r/evm` lockfile entry still resolves with `@x402/evm` + `viem` in its peer suffix — that's expected since those are `@x402r/evm`'s peers being satisfied by the broader workspace. Nothing was dropped from resolution, just from `helpers`' explicit declaration.

**devDependency pin for `@x402r/evm: 0.2.0-alpha.1`**
Exact lower-bound pin is intentional and correct — matches the peer range floor `>=0.2.0-alpha.1 <0.3.0`. ✓

**Changeset**
Correctly tagged `patch`. Description accurately describes the consumer-visible impact (peer surface unchanged, just re-routed through `@x402r/evm`).

---

### Nits

- **Changeset body** is slightly inconsistent with the PR description: it says "only requires `@x402/core` and `@x402r/evm`" but omits mentioning that `@x402/core` is listed in peers as `@x402/core` (without the `r`). Not wrong, just worth keeping that distinction clear in any follow-up docs since the `@x402/core` vs `@x402r/core` naming is already a point of confusion in this ecosystem.

- **`devDependencies` still has no `@x402/core` pin** — wait, it does: `"@x402/core": "2.14.0"` (already present, unchanged). Ignore this. ✓

- Minor: The PR body says "No internal workspace package depends on `@x402r/helpers`" — worth a quick `pnpm why @x402r/helpers` confirmation if that assumption ever changes, since any internal consumer would inherit the narrowed peer surface without a semver bump. Could be a comment in `package.json` or just tracked as a follow-up.

---

**Verdict:** approve. Peer hygiene is correct, lockfile is clean, tests pass. No behavior change for consumers.

tofu-miso (commented):
## Review: chore(helpers): drop unused @x402/evm + viem peer dependencies

**LGTM overall — clean, well-validated housekeeping.** A few observations below.

---

### Overview

Removes `@x402/evm` and `viem` from `@x402r/helpers`' `peerDependencies` and `devDependencies` since neither is directly imported by the package. Peer exposure is preserved transitively via `@x402r/evm`'s own peer declarations. Changeset, lockfile, and tests are all in order.

---

### What's Good

- **Rationale is correct.** The package only imports `@x402r/evm` (the `r` variant) and `@x402/core/server` — declaring `@x402/evm` + `viem` as peers was a false claim about runtime requirements.
- **Transitive peer reasoning is sound.** `@x402r/evm@0.2.0-alpha.1` already pulls in `@x402/evm` + `viem` as its own peers, so consumers aren't left with an unmet peer they can't find.
- **Validation is thorough.** 30/30 tests, typecheck, build, publint/attw, knip, and frozen-lockfile install — good coverage for a dependency change.
- **Lockfile diff is surgical.** Only the helpers importer's entries are removed; nothing else shifted.
- **Changeset is correctly scoped as `patch`.** Peer dep removal with no behavioral change → patch is right.

---

### Nits / Minor Observations

**1. Transitive peer coupling is now load-bearing**

The cleanup is correct today, but it creates an implicit assumption: `@x402r/evm` will always re-export `viem` + `@x402/evm` as peers. If a future version of `@x402r/evm` internalises either dep or drops the peer declaration, consumers of `@x402r/helpers` would lose the signal without a breaking change here. Worth a short comment in `package.json` or the ADR/docs tracking peer strategy, not a code change.

**2. Alpha version range in peerDeps**

`"@x402r/evm": ">=0.2.0-alpha.1 <0.3.0"` — the open upper bound to `<0.3.0` is fine, but alpha semver often isn't respected strictly. If `0.2.0-alpha.2` changes the peer surface, the range gives consumers false confidence. Already pre-existing so not a blocker here, but worth tracking.

**3. `devDependencies` pin vs. peer range**

`devDependencies` pins `@x402r/evm` at exactly `0.2.0-alpha.1` while peer declares `>=0.2.0-alpha.1 <0.3.0`. Minor but common — consider pinning devDep to the floor of the range (`0.2.0-alpha.1` is fine) so CI tests the minimum-supported version. Already the case here, so no action needed unless you cut `0.2.0-alpha.2`.

---

### Verdict

No blocking issues. The cleanup is correct and well-tested. The nits are informational — nothing that should hold this up.

tofu-miso (approved):
LGTM — nits are informational only, no blockers.

vraspar:
Thanks @tofu-miso — these are all on the (intended) peer-removal design and non-blocking, so no code changes here:

- **viem / `@x402/evm` reachable only via `@x402r/evm`'s peers:** that's the intent. `@x402r/helpers` requires `@x402r/evm`, which declares `viem` + `@x402/evm` as its *own* peerDependencies — so a consumer still surfaces them, just through `@x402r/evm` rather than a redundant re-declaration here. (A broader peer-strategy note could be a separate docs follow-up if useful.)
- **`>=0.2.0-alpha.1 <0.3.0` range strictness:** unchanged by this PR — that range was set by an earlier changeset; this PR only removes the `viem` + `@x402/evm` peers.
- **devDep pinned to the range floor:** already the case (`@x402r/evm@0.2.0-alpha.1`).
- **`pnpm why @x402r/helpers` (no internal consumers):** confirmed — zero internal workspace consumers, so this affects external npm consumers only, and their required-peer surface is unchanged.
- **changeset `@x402/core` vs `@x402r/core` naming:** both are spelled exactly as `package.json` lists them (`@x402/core` is the upstream package; `@x402r/evm`/`@x402r/core` are ours) — accurate as written.

Happy to open a separate peer-strategy doc note if that'd help, but nothing to change in this PR.

tofu-miso (commented):
Thanks for the detailed responses — all nits are resolved:

- **Transitive peer coupling:** understood, the design is intentional. `@x402r/evm` owns the `viem` / `@x402/evm` peer declarations so consumers still surface them — the re-declaration here was genuinely redundant.
- **Alpha range:** noted, pre-existing from an earlier changeset, not in scope here.
- **DevDep floor pin:** already aligned ✓
- **No internal consumers:** `pnpm why` confirms it — external consumers only, peer surface unchanged.
- **`@x402/core` naming:** accurate as written ✓

Nothing left to address. Original approval stands — ship it.

A1igator:
### SDK Review

Found 0 issues (axes: tests, conventions, dead code).

Verified at d5c9550: no direct `viem` or `@x402/evm` imports anywhere in `packages/helpers` src or tests (the only `viem` mention is a design comment in `x402r-defaults.ts`); `@x402r/evm@0.2.0-alpha.1` declares `viem` and `@x402/evm` as its own peers (confirmed via npm registry), so the consumer peer surface is unchanged as the changeset claims; no stale references to the dropped deps in the README, tsdown/vitest/tsconfig configs, or size-limit entries; lockfile removal is consistent with the new `package.json`.

---

A1igator (approved):
Approving per review-sdk: 0 issues across tests, conventions, and dead code axes. Verification details in the review comment above.

PR #210, opened 2026-06-12
refactor(core): source payment hash from on-chain getHash in getPaymentState

vraspar:
## What

`getPaymentState` now reads the escrow's own on-chain `getHash(paymentInfo)` view to key `paymentState`, instead of recomputing that hash locally with the TS `computePaymentInfoHash`.

## Why

`getHash` is the exact contract function the local TS hash mirrors, so it returns the identical `paymentState` key — but reading it from the chain means it **can never drift** from the contract (the source of truth). This is the same on-chain-truth principle the SDK already favors. It also makes `getPaymentState` no longer a consumer of `computePaymentInfoHash`, which lets that redundant TS hash be removed in a follow-up.

## Behavior & tradeoff

Public signature and return value are unchanged. The one observable difference: `getState`/`getAmounts` now issue an extra `eth_call` per invocation — the reads are inherently sequential (`ESCROW` → `getHash` → `paymentState`, each depends on the prior), so they can't be collapsed into a multicall. That RPC-count change is consumer-visible (mocks, call-count assertions, metered providers), so a patch changeset is included.

## Tests

- `packages/core/tests/payment-state.test.ts` mock rebased from 2 reads → 3; the old "different chainId → different hash" unit test is reframed to assert the hash flows `getHash → paymentState` (chainId no longer drives the hash locally).
- `pnpm --filter @x402r/core test:unit`: 226/226. `pnpm typecheck`: 6/6.
- The real `getHash` RPC path is also exercised by the existing fork tests that call `getAmounts`.

## Scope note

`computePaymentInfoHash` is **intentionally kept** here — it still has other consumers. Its removal is a separate follow-up PR (the payer-signing removal). This PR only stops `getPaymentState` from using it.

Part of the authCapture upstream-alignment Phase 3 work.

### DIFF
--- a/.changeset/core-getpaymentstate-onchain-gethash.md
+++ b/.changeset/core-getpaymentstate-onchain-gethash.md
@@ -1,0 +1,5 @@
+---
+"@x402r/core": patch
+---
+
+`getPaymentState` (`getState`/`getAmounts`) now reads the payment hash from the escrow's on-chain `getHash` instead of computing it locally, adding one RPC read per call.
--- a/packages/core/src/actions/escrow/getPaymentState.ts
+++ b/packages/core/src/actions/escrow/getPaymentState.ts
@@ -3,7 +3,6 @@ import {
   authCaptureEscrowAbi,
   paymentOperatorAbi,
 } from '../../abis/generated.js'
-import { computePaymentInfoHash } from '../../payment/hashing.js'
 import type { PaymentInfo } from '../../types/index.js'
 import { wrapContractCall } from '../_internal/error-wrapping.js'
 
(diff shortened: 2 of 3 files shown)


vraspar:
Closing — decided against the on-chain getHash switch. The drift argument doesn't justify the cost: if the Commerce Payments AuthCaptureEscrow hashing ever changed, that'd be a breaking protocol redeployment that surfaces far louder than a silent TS hash mismatch, so local computePaymentInfoHash drift isn't a realistic standalone failure mode — and an extra RPC read per getState/getAmounts isn't worth that marginal safety. The scheme-side hash copy is also moving upstream into x402, easing the dedup pressure. Keeping getPaymentState computing the hash locally.

PR #215, opened 2026-06-12
refactor(examples): build the http-wire-capture extra with x402rDefaults

A1igator:
## Summary

The http-wire-capture scenario was the only place in the repo that builds an auth-capture `extra` by hand, and four of its eight literals (`minFeeBps: 0`, `maxFeeBps: 100`, `name: 'USDC'`, `version: '2'`) are exactly the defaults `x402rDefaults` ships. The SDK was not dogfooding its own helper.

This routes the scenario through `x402rDefaults`, keeping the scenario-specific values explicit:

- `captureAuthorizer` (required input)
- absolute `captureDeadline` / `refundDeadline` derived from the Anvil block timestamp (the helper's opt-in absolute escape hatch)
- `feeRecipient: zeroAddress` (deferred fee-recipient selection; the helper would otherwise default it to the captureAuthorizer)

Also adds `@x402r/helpers` as a workspace dep of `examples`.

No changeset: only the private `examples` package changes.

## Test plan

- [x] `pnpm typecheck` clean (7/7 tasks)
- [x] `pnpm check` (biome) clean
- [x] `pnpm --filter examples run scenario:http-wire-capture` passes end-to-end against an Anvil fork of Base Sepolia (6 steps, all post-conditions green, including the PaymentAuthorized event and escrow-state assertions that would catch any drift in the produced `extra`)

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@x402r/core": "workspace:*",
+    "@x402r/helpers": "workspace:*",
     "@x402r/sdk": "workspace:*",
     "prool": "^0.2.4",
     "tsx": "^4.22.4",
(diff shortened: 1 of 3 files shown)


A1igator:
Closing: keeping example changes scoped to arbiter-examples PR 13 for now. The dogfooding idea stays tracked in the issues if we want it later.

